PR #29, opened 2025-10-23
chore: API Server - Permissions Error returns a 403 response

camielvs:
API will now return a 403 response when a user is not authorized to take an action. e.g. cancel a pipeline.

### DIFF
--- a/api_server_main.py
+++ b/api_server_main.py
@@ -4,6 +4,7 @@ import traceback
 import fastapi
 
 from cloud_pipelines_backend import api_router
+from cloud_pipelines_backend import database_ops
 
 app = fastapi.FastAPI(
     title="Cloud Pipelines API",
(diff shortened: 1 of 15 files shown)


cloud_pipelines_backend/errors.py
```diff
@@ -7,4 +7,7 @@ class ItemAlreadyExistsError(Exception):
 
 
 class PermissionError(Exception):
-    pass
+    def __init__(self, message: str):
```
Ark-kun: This might be unnecessary. The base `Exception` class already takes arguments. `str(exc)` returns a representation of those arguments. And if there was only one argument (e.g. message), then `str(exc)` will equal that argument.

```
class PermissionError(Exception):
    pass

ex = PermissionError("some text")                 
print(str(ex))
# >> some text
```

As for status code I don't think it's not appropriate here. These exceptions are raised in non-HTTP code (api_server), so HTTP status codes are not applicable.

The HTTP Server layer (api_router) then converts those business-logic exceptions to HTTP exceptions and that's when status HTTP codes appear.

Ark-kun (approved):

cloud_pipelines_backend/errors.py
camielvs: Correct! I've updated - now it follows the same pattern as the 404 error.

PR #65, opened 2026-01-08
refactor: Centralize CORS business logic into Tangle middleware

morgan-wowk:
**Why is this change needed:**

1. It centralizes the CORS logic for Tangle so we are not maintaining duplicate solutions.
2. It introduces the ability to configure CORS, increasing support for various environment configuration:
    - Local development (frontend on `localhost:3000`, backend on `localhost:8000`)
    - Production deployments where the frontend and backend are on different domains
    - Multi-tenant deployments with multiple frontend URLs

### DIFF
--- a/cloud_pipelines_backend/middleware.py
+++ b/cloud_pipelines_backend/middleware.py
@@ -1,0 +1,94 @@
+"""
+Global middleware configuration for Tangle API servers.
+
+This module provides reusable middleware setup functions that should be used
+by all entry points to ensure consistent behavior across deployments.
+"""
+
+import logging
+import os
+from urllib.parse import urlparse
+
+import fastapi
+from fastapi.middleware.cors import CORSMiddleware
+
+
+logger = logging.getLogger(__name__)
+
+
+def _is_valid_origin(origin: str) -> bool:
+    """
+    Validate that an origin string is a valid URL format.
+
+    Args:
+        origin: The origin URL to validate
+
+    Returns:
+        True if valid, False otherwise
+    """
+    try:
+        parsed = urlparse(origin)
+        # Must have a scheme (http/https) and a netloc (domain/host)
+        if not parsed.scheme or not parsed.netloc:
+            return False
+        # Scheme must be http or https
+        if parsed.scheme not in ("http", "https"):
+            return False
+        # Should not have a path beyond '/'
+        if parsed.path and parsed.path != "/":
+            return False
+        return True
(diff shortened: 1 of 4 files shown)


cloud_pipelines_backend/middleware.py
```diff
@@ -0,0 +1,72 @@
...
+    Environment Variables:
+        TANGLE_CORS_ALLOWED_ORIGINS: Comma-separated list of allowed origins.
+            Default: "http://localhost:3000,http://127.0.0.1:3000" if not set.
+    """
+    cors_origins_str = os.environ.get("TANGLE_CORS_ALLOWED_ORIGINS", "http://localhost:3000,http://127.0.0.1:3000")
+    
+    # Parse the comma-separated list and strip whitespace from each origin
+    allowed_origins = [
+        origin.strip() 
+        for origin in cors_origins_str.split(",")
```
maxy-shpfy: Do we want to add more sanity-checks (aka valid URLs) since it is a user-input?

cloud_pipelines_backend/middleware.py
```diff
@@ -0,0 +1,72 @@
...
+        allow_headers=["*"],
+    )
+    
+    logger.info(
+        f"CORS middleware configured for {len(allowed_origins)} origin(s): "
+        f"{', '.join(allowed_origins)}"
+    )
+
+
+def setup_global_middleware(app: fastapi.FastAPI) -> None:
```
maxy-shpfy: NIT: No idea what naming conventions usually are, but since the usage is `middleware.setup_global_middleware` , maybe we simplify to `middlware.setup`? Not sure what not "global" middleware we may have.

cloud_pipelines_backend/middleware.py
morgan-wowk: Great idea. I am sure this will help some people.

I have added some validation and useful messages when invalid formats are detected.

cloud_pipelines_backend/middleware.py
morgan-wowk: I think `setup` is fine for now as well. Thank you!

Until we have a group of middleware to call, I am just calling `setup_cors_middleware` directly now (1 less function).

morgan-wowk:
> LGTM, but havent had a chance to test

Thanks for reviewing. I have added tests to the PR now as well to verify the middleware is meeting expectations, then will test again on staging with actual requests.

PR #66, opened 2026-01-12
feat: Introduce logging context to Tangle

morgan-wowk:
**Issue**
https://github.com/TangleML/tangle/issues/67

**Changes:**

* Adds trace context helpers
* Add trace middleware to apply trace context to API requests
    * This helps logging formatters pull trace ids out of context to set them in logs
    * Sets the trace id on the response for client consumption
* Wraps API requests with trace context using middleware
* Passes trace id to pipeline executions using `extras` parameter
* Wraps pipeline execution with trace context (using trace id from `extras` parameter, or creating a new one if None)

**Test pipeline notes:**

* Requires `ytest`, `starlette`, `httpx` (e.g. `pip3 install ...`)

### DIFF
--- a/cloud_pipelines_backend/instrumentation/api_tracing.py
+++ b/cloud_pipelines_backend/instrumentation/api_tracing.py
@@ -1,0 +1,61 @@
+"""Request context middleware for FastAPI applications.
+
+This middleware automatically generates a request_id for each incoming HTTP request,
+sets it in the logging context for the duration of the request, and includes it in
+the response headers.
+"""
+
+import logging
+import secrets
+
+from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.requests import Request
+from starlette.responses import Response
+
+from . import contextual_logging
+
+logger = logging.getLogger(__name__)
+
+
+def generate_request_id() -> str:
+    """Generate a new request ID compatible with OpenTelemetry format.
+
+    OpenTelemetry trace IDs are 16-byte (128-bit) values represented as
+    32 hexadecimal characters (lowercase). We use the same format for
+    request IDs to maintain compatibility.
+
+    Returns:
+        A 32-character hexadecimal string representing the request ID
+    """
+    return secrets.token_hex(16)
+
+
+class RequestContextMiddleware(BaseHTTPMiddleware):
+    """Middleware to manage request_id for each request.
+
+    For each incoming request:
+    1. Generates a new request_id (32-character hex string)
+    2. Sets it in the logging context (as 'request_id' key)
+    3. Adds it to the response headers as 'x-tangle-request-id'
+    4. Clears it after the request completes
+
+    This ensures all logs during the request processing include the same request_id.
+    """
+
+    async def dispatch(self, request: Request, call_next) -> Response:
+        """Process each request with a new request_id.
+
+        Args:
+            request: The incoming HTTP request
+            call_next: The next middleware or route handler
+
+        Returns:
+            The HTTP response with request_id in headers
+        """
+        # Generate a new request_id for this request
+        request_id = generate_request_id()
+
+        with contextual_logging.logging_context(request_id=request_id):
+            response = await call_next(request)
+            response.headers["x-tangle-request-id"] = request_id
+            return response
--- a/cloud_pipelines_backend/instrumentation/contextual_logging.py
+++ b/cloud_pipelines_backend/instrumentation/contextual_logging.py
@@ -1,0 +1,127 @@
+"""Logging context management for distributed tracing and execution tracking.
+
+This module provides utilities for managing arbitrary metadata in the logging context.
+This metadata is automatically added to all log records for better filtering and correlation.
+
+Common metadata keys:
+- request_id: From API requests - groups all logs from a single API call
+- pipeline_run_id: From PipelineRun.id - tracks the entire pipeline run
+- execution_id: From ExecutionNode.id - tracks individual execution nodes
+- container_execution_id: From ContainerExecution.id - tracks running containers
+- user_id: User who initiated the operation
+- Any other metadata you want to track in logs
+
+Usage:
+    # Set metadata in context
+    with logging_context(request_id="abc123", user_id="user@example.com"):
+        logger.info("Processing")  # Both fields in logs
+
+    # Or use individual functions
+    set_context_metadata("request_id", "abc123")
+    delete_context_metadata("request_id")  # Remove a specific key
+"""
+
+import contextvars
+from contextlib import contextmanager
+from typing import Any, Optional
+
+# Single context variable to store all metadata as a dictionary
+_context_metadata: contextvars.ContextVar[dict[str, Any]] = contextvars.ContextVar(
+    "context_metadata", default={}
+)
+
+
+def set_context_metadata(key: str, value: Any) -> None:
+    """Set a metadata value in the current context.
+
+    Args:
+        key: The metadata key (e.g., 'execution_id', 'request_id', 'user_id')
+        value: The value to set
+    """
+    metadata = _context_metadata.get().copy()
+    metadata[key] = value
+    _context_metadata.set(metadata)
+
+
+def delete_context_metadata(key: str) -> None:
+    """Delete a metadata key from the current context.
+
+    Similar to dict.pop() but doesn't return a value. If the key doesn't exist,
+    this is a no-op (no error is raised).
+
+    Args:
+        key: The metadata key to delete (e.g., 'execution_id', 'request_id')
+    """
+    metadata = _context_metadata.get().copy()
+    metadata.pop(key, None)  # Use None as default to avoid KeyError
+    _context_metadata.set(metadata)
+
+
+def get_context_metadata(key: str) -> Optional[Any]:
+    """Get a metadata value from the current context.
+
+    Args:
+        key: The metadata key to retrieve
+
+    Returns:
+        The metadata value or None if not set
+    """
+    return _context_metadata.get().get(key)
+
+
+def get_all_context_metadata() -> dict[str, Any]:
+    """Get all metadata from the current context.
+
+    Returns:
+        Dictionary of all context metadata
+    """
+    return _context_metadata.get().copy()
+
+
+def clear_context_metadata() -> None:
+    """Clear all metadata from the current context."""
+    _context_metadata.set({})
+
+
+@contextmanager
+def logging_context(**metadata: Any):
+    """Context manager for setting arbitrary metadata that is automatically cleared.
+
+    This is the recommended way to set logging context. It ensures metadata is
+    always cleaned up, even if an exception occurs.
+
+    You can pass any keyword arguments, and they will be available in log records.
+    Common keys include: request_id, pipeline_run_id, execution_id, container_execution_id, user_id
+
+    Args:
+        **metadata: Arbitrary keyword arguments to add to the context
+
+    Example with IDs:
+        >>> with logging_context(pipeline_run_id="run123", execution_id="exec456"):
+        ...     logger.info("Processing execution")  # Will include both IDs
+
+    Example with custom metadata:
+        >>> with logging_context(
+        ...     execution_id="exec456",
+        ...     user_id="user@example.com",
+        ...     operation="reprocessing"
+        ... ):
+        ...     logger.info("Custom operation")  # All metadata in logs
+
+    Example for API requests:
+        >>> request_id = generate_request_id()
+        >>> with logging_context(request_id=request_id):
+        ...     logger.info("Handling API request")
+    """
+    # Store previous metadata to restore nested contexts
+    prev_metadata = get_all_context_metadata()
+
+    try:
+        # Set all provided metadata
+        for key, value in metadata.items():
+            if value is not None:  # Only set non-None values
+                set_context_metadata(key, value)
+        yield
+    finally:
+        # Restore previous metadata
+        _context_metadata.set(prev_metadata)
--- a/cloud_pipelines_backend/orchestrator_sql.py
+++ b/cloud_pipelines_backend/orchestrator_sql.py
@@ -20,7 +20,7 @@ from . import backend_types_sql as bts
 from . import component_structures as structures
 from .launchers import common_annotations
 from .launchers import interfaces as launcher_interfaces
-
+from .instrumentation import contextual_logging
 
 _logger = logging.getLogger(__name__)
 
@@ -95,24 +95,29 @@ class OrchestratorService_Sql:
         if queued_execution:
             self._queued_executions_queue_idle = False
             start_timestamp = time.monotonic_ns()
-            _logger.info(f"Before processing {queued_execution.id=}")
-            try:
-                self.internal_process_one_queued_execution(
-                    session=session, execution=queued_execution
-                )
-            except Exception as ex:
-                _logger.exception(f"Error processing {queued_execution.id=}")
-                session.rollback()
-                queued_execution.container_execution_status = (
-                    bts.ContainerExecutionStatus.SYSTEM_ERROR
-                )
-                record_system_error_exception(execution=queued_execution, exception=ex)
-                session.commit()
-            finally:
-                duration_ms = int((time.monotonic_ns() - start_timestamp) / 1_000_000)
-                _logger.info(
-                    f"After processing {queued_execution.id=}. Duration={duration_ms}ms"
-                )
+
+            with contextual_logging.logging_context(execution_id=queued_execution.id):
+                _logger.info("Before processing queued execution")
+                try:
+                    self.internal_process_one_queued_execution(
+                        session=session, execution=queued_execution
+                    )
+                except Exception as ex:
+                    _logger.exception("Error processing queued execution")
+                    session.rollback()
+                    queued_execution.container_execution_status = (
+                        bts.ContainerExecutionStatus.SYSTEM_ERROR
+                    )
+                    record_system_error_exception(
+                        execution=queued_execution, exception=ex
+                    )
+                    session.commit()
+                finally:
+                    duration_ms = (time.monotonic_ns() - start_timestamp) / 1_000_000
+                    _logger.info(
+                        f"After processing queued execution (duration: {duration_ms}ms)"
+                    )
+
             return True
         else:
             if not self._queued_executions_queue_idle:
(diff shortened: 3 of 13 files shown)


cloud_pipelines_backend/instrumentation/trace_middleware.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        # Generate a new trace_id for this request
+        trace_id = trace_context.generate_trace_id()
+        trace_context.set_trace_id(trace_id)
+        
+        try:
+            # Process the request
+            response = await call_next(request)
+            # Add trace_id to response headers for client reference
+            response.headers["x-tangle-trace-id"] = trace_id
```
morgan-wowk: Note to self. This line should probably go above `await call_next(request)` in case an exception is raised

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -94,20 +96,36 @@ def internal_process_queued_executions_queue(self, session: orm.Session):
...
-            except Exception as ex:
-                _logger.exception(f"Error processing {queued_execution.id=}")
-                session.rollback()
-                queued_execution.container_execution_status = (
-                    bts.ContainerExecutionStatus.SYSTEM_ERROR
+            
+            # Restore trace_id from execution
+            trace_id = None
+            if queued_execution.extra_data:
+                trace_id = queued_execution.extra_data.get("trace_id")
```
Ark-kun: I do not think the request trace IDs are relevant when processing execution nodes. Execution nodes already have their own IDs and they might work better as trace IDs.

morgan-wowk:
> Thank you for this PR.
> 
> I'm reviewing the changes and I think we might want to design this slightly differently. I think that trace IDs are mostly relevant to API Server and API requests. Orchestrator works autonomously, it's not triggered by API requests, so it does not have them or need them. Instead, the ExecutionNode.id and ContainerExecution.id already serve the same role.
> 
> On the other hand, API requests might benefit from trace IDs so that all logging messages that are generated when processing a single API call can be filtered and grouped together.

This sounds good to me.

I will make some adjustments. Thanks!

morgan-wowk:
> Thank you for this PR.
> 
> I'm reviewing the changes and I think we might want to design this slightly differently. I think that trace IDs are mostly relevant to API Server and API requests. Orchestrator works autonomously, it's not triggered by API requests, so it does not have them or need them. Instead, the ExecutionNode.id and ContainerExecution.id already serve the same role.
> 
> On the other hand, API requests might benefit from trace IDs so that all logging messages that are generated when processing a single API call can be filtered and grouped together.

The work has been re-designed. Here is a summary:

# Logging Context Improvements

## Summary

Enhanced the Tangle orchestrator logging to automatically include execution context (execution_id, container_execution_id) in all log messages, making it easier to trace and filter logs for specific executions.

## Changes Made

### 1. Added Logging Context to Orchestrator (`orchestrator_sql.py`)

Wrapped execution processing with `logging_context.logging_context()` to automatically attach execution IDs to all logs within that scope.

#### For Queued Executions:

```python
# Set execution context for logging
with logging_context.logging_context(
    execution_id=queued_execution.id
):
    _logger.info("Before processing queued execution")
    # ... process execution ...
    _logger.info("After processing queued execution")
```

#### For Running Container Executions:

```python
with logging_context.logging_context(
    execution_id=execution_id,
    container_execution_id=running_container_execution.id
):
    _logger.info("Before processing running container execution")
    # ... process container execution ...
    _logger.info("After processing running container execution")
```

### 2. Added Context-Aware Logging Configuration (`start_local.py`)

Created a custom formatter and filter that dynamically includes context fields in logs only when they're set.

```python
class LoggingContextFilter(logging.Filter):
    """Adds contextual metadata to log records."""
    
    def filter(self, record: logging.LogRecord) -> bool:
        for key, value in get_all_context_metadata().items():
            if value is not None:
                setattr(record, key, value)
        return True


class ContextAwareFormatter(logging.Formatter):
    """Formatter that dynamically includes context fields only when they're set."""
    
    def format(self, record: logging.LogRecord) -> str:
        base_format = "%(asctime)s [%(levelname)s] %(name)s"
        
        # Collect context fields that are present
        context_parts = []
        context_metadata = get_all_context_metadata()
        for key, value in context_metadata.items():
            if value is not None and hasattr(record, key):
                context_parts.append(f"{key}={value}")
        
        # Add context to format if any exists
        if context_parts:
            base_format += " [" + " ".join(context_parts) + "]"
        
        base_format += ": %(message)s"
        
        formatter = logging.Formatter(base_format)
        return formatter.format(record)
```

### 3. Cleaned Up Redundant IDs in Log Messages

Removed execution IDs from log messages where they're already present in the logging context, eliminating duplication.

**Examples of cleaned up logs:**

```python
# Before: ID appears twice (in context and message)
_logger.info(f"Before processing {queued_execution.id=}")

# After: ID only in context
_logger.info("Before processing queued execution")
```

```python
# Before
_logger.info(f"Container execution {container_execution.id} is now in state {new_status}")

# After
_logger.info(f"Container execution is now in state {new_status}")
```

```python
# Before
_logger.info(f"Terminating container execution {container_execution.id}.")

# After
_logger.info("Terminating container execution.")
```

## Log Output Examples

### Without Context (non-orchestrator logs)

```
2026-01-19 14:40:52,611 [INFO] start_local: Starting the orchestrator
2026-01-19 14:40:52,620 [INFO] uvicorn.error: Application startup complete.
```

### With Execution Context Only

```
2026-01-19 14:30:13,650 [INFO] cloud_pipelines_backend.orchestrator_sql [execution_id=019bd8612bb80ea96aba]: Before processing queued execution
2026-01-19 14:30:13,674 [INFO] cloud_pipelines_backend.orchestrator_sql [execution_id=019bd8612bb80ea96aba]: Execution will reuse the old_execution.id='019b3428a0f1e29c82d4'
2026-01-19 14:30:13,681 [INFO] cloud_pipelines_backend.orchestrator_sql [execution_id=019bd8612bb80ea96aba]: After processing queued execution
```

### With Both Execution and Container Execution Context

```
2026-01-19 14:30:14,687 [INFO] cloud_pipelines_backend.orchestrator_sql [execution_id=019bd8612bb888cc61d4 container_execution_id=019b342ab120726fdee2]: Before processing running container execution
2026-01-19 14:30:14,701 [INFO] cloud_pipelines_backend.orchestrator_sql [execution_id=019bd8612bb888cc61d4 container_execution_id=019b342ab120726fdee2]: Container execution is now in state RUNNING (was PENDING)
2026-01-19 14:30:14,713 [INFO] cloud_pipelines_backend.orchestrator_sql [execution_id=019bd8612bb888cc61d4 container_execution_id=019b342ab120726fdee2]: After processing running container execution
```

## Benefits

1. **Easy Log Filtering**: Filter logs by execution_id or container_execution_id in log aggregation tools
2. **Better Traceability**: Follow an execution through its entire lifecycle
3. **No Duplication**: IDs appear once in context, not repeated in every message
4. **Dynamic Context**: Formatter automatically includes any context fields without code changes
5. **Cleaner Messages**: Log messages are more readable without ID clutter

## Extensibility

The context system is fully extensible. You can add any arbitrary metadata:

```python
with logging_context.logging_context(
    execution_id=execution.id,
    container_execution_id=container.id,
    pipeline_run_id=pipeline.id,  # Would work if added
    user_id=user.id,               # Would work if added
    custom_field="value"           # Any custom field works
):
    _logger.info("Processing with custom context")
```

All fields automatically appear in logs:

```
[INFO] module [execution_id=123 container_execution_id=456 pipeline_run_id=789 user_id=admin custom_field=value]: Processing with custom context
```

## Files Modified

- `tangle/cloud_pipelines_backend/orchestrator_sql.py` - Added logging context to execution processing
- `tangle/start_local.py` - Added `LoggingContextFilter` and `ContextAwareFormatter`

## Backward Compatibility

- ✅ All changes are backward compatible
- ✅ Logs without context still work (no context fields shown)
- ✅ Existing log parsing tools continue to work
- ✅ All tests pass

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -94,20 +96,26 @@ def internal_process_queued_executions_queue(self, session: orm.Session):
...
-                )
-                record_system_error_exception(execution=queued_execution, exception=ex)
-                session.commit()
-            _logger.info(f"After processing {queued_execution.id=}")
+            
+            # Set execution context for logging
+            with logging_context.logging_context(
+                execution_id=queued_execution.id
+            ):
+                _logger.info("Before processing queued execution")
```
Ark-kun, line 100: Will the execution ID information still be present in non-JSON logs?

morgan-wowk: It will be, yes. This is achieved through the following change:

```python

class ContextAwareFormatter(logging.Formatter):
    """Formatter that dynamically includes context fields only when they're set."""
    
    def format(self, record: logging.LogRecord) -> str:
        """Format log record with dynamic context fields."""
        # Base format
        base_format = "%(asctime)s [%(levelname)s] %(name)s"
        
        # Collect context fields that are present
        context_parts = []
        context_metadata = get_all_context_metadata()
        for key, value in context_metadata.items():
            if value is not None and hasattr(record, key):
                context_parts.append(f"{key}={value}")
        
        # Add context to format if any exists
        if context_parts:
            base_format += " [" + " ".join(context_parts) + "]"
        
        base_format += ": %(message)s"
        
        # Create formatter with the dynamic format
        formatter = logging.Formatter(base_format)
        return formatter.format(record)
```

start_local.py
```diff
@@ -78,19 +78,70 @@ def get_user_details(request: fastapi.Request):
 
 # region: Logging configuration
 import logging.config
+from cloud_pipelines_backend.instrumentation.logging_context import get_all_context_metadata
+
+class LoggingContextFilter(logging.Filter):
```
morgan-wowk: Move these two classes to a dedicated, well-named file/module rather than being coupled with an entrypoint script.

cloud_pipelines_backend/orchestrator_sql.py
Ark-kun: Thank you. I wonder how `_logger.exception` messages look like.

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -132,37 +140,46 @@ def internal_process_running_executions_queue(self, session: orm.Session):
...
-                    record_system_error_exception(
-                        execution=execution_node, exception=ex
+            
+            # Set execution context for logging (includes container_execution_id)
+            # Get first execution_id for context (there may be multiple nodes using same container)
+            execution_nodes = running_container_execution.execution_nodes
+            execution_id = execution_nodes[0].id if execution_nodes else None
+            
+            with logging_context.logging_context(
+                execution_id=execution_id,
```
Ark-kun: Let's call it: `execution_node_id`

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -276,7 +293,7 @@ def internal_process_one_queued_execution(
             # There must be at least one SUCCEEDED/RUNNING/PENDING since non_purged_candidates is non-empty.
             old_execution = non_purged_candidates[-1]
             _logger.info(
-                f"Execution {execution.id=} will reuse the {old_execution.id=} with "
```
Ark-kun, line 289: Let's either restore the `=` or rephrase: `Reusing cached execution node {old_execution.id} with`

start_local.py
```diff
@@ -78,19 +78,70 @@ def get_user_details(request: fastapi.Request):
 
 # region: Logging configuration
 import logging.config
+from cloud_pipelines_backend.instrumentation.logging_context import get_all_context_metadata
```
Ark-kun: style: Let's only import modules, not classes/function. https://google.github.io/styleguide/pyguide.html#22-imports

```
import foo
from foo import bar
```

If the module name is ambiguous/conflicting, tehn it can be renamed:

```
from launchers import interfaces as launcher_interfaces
```

cloud_pipelines_backend/instrumentation/contextual_logging.py
Ark-kun, line 1: Ideally, module names should be plural (e.g. "contexts") or gerund ("logging").
Singular nouns clash with local variables. For example, Kubernetes library has `from kubernetes import client` and it's a pain.

In this case it's harder to find a good name.
Maybe `contextual_logging.py`?

cloud_pipelines_backend/instrumentation/api_tracing.py
Ark-kun, line 1: Maybe, `api_tracing.py`?

cloud_pipelines_backend/orchestrator_sql.py
Ark-kun, line 1: Let's make sure to use Black formatting on all modified files.

cloud_pipelines_backend/instrumentation/contextual_logging.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def set_context_metadata(key: str, value: Any) -> None:
+    """Set a metadata value in the current context.
+    
+    Args:
+        key: The metadata key (e.g., 'execution_id', 'request_id', 'user_id')
+        value: The value to set
+    """
+    metadata = _context_metadata.get().copy()
```
Ark-kun, line 41: What is the main purpose of `.copy()` here?

cloud_pipelines_backend/instrumentation/contextual_logging.py
```diff
@@ -0,0 +1,111 @@
...
+from typing import Any, Optional
+
+
+# Single context variable to store all metadata as a dictionary
+_context_metadata: contextvars.ContextVar[dict[str, Any]] = contextvars.ContextVar(
+    "context_metadata", default={}
+)
+
+
+def set_context_metadata(key: str, value: Any) -> None:
```
Ark-kun, line 34: Is there a way to delete/unset a key?
Similar to a dict.

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: I tested this and it outputs like this:

```
2026-01-22 10:13:10,839 [ERROR] test_logger [execution_id=exec-99999]: Error logged with .exception()
Traceback (most recent call last):
  File "/Users/morganwowk/src/github.com/morgan-wowk/tangleml-docker/tangle/tests/test_context_aware_formatter_exception.py", line 185, in test_logger_exception_with_context_aware_formatter
    raise ValueError("This is a test error")
ValueError: This is a test error
```

`test_logger` is the name of the logger and I'm not sure that's valuable 🤔 

We can remove that portion of the format in a follow-up PR if you also don't like it.

cloud_pipelines_backend/instrumentation/contextual_logging.py
morgan-wowk: ## Why `.copy()` is Critical Here

The `.copy()` on line 37 is **critical for context isolation** when working with `contextvars`.

### The Purpose

The `.copy()` ensures that each context has its own independent dictionary. Without it, you'd be mutating a shared dictionary reference, which would break context isolation between:
- Different async tasks
- Different threads  
- Nested contexts

### Why This Matters

```python
# WITHOUT .copy() (WRONG):
metadata = _context_metadata.get()  # Gets reference to dict
metadata[key] = value  # Mutates the SHARED dict
_context_metadata.set(metadata)  # All contexts see this change!

# WITH .copy() (CORRECT):
metadata = _context_metadata.get().copy()  # Gets NEW dict
metadata[key] = value  # Mutates ONLY our copy
_context_metadata.set(metadata)  # Only our context sees this
```

### Concrete Example

**Without `.copy()`:**
```python
# Task 1 sets request_id
await task1()  # Sets request_id="abc"

# Task 2 sets request_id  
await task2()  # Sets request_id="xyz"

# Task 1's logs would incorrectly show request_id="xyz"! ❌
```

**With `.copy()`:**
```python
# Task 1 sets request_id
await task1()  # Sets request_id="abc" in its own copy

# Task 2 sets request_id  
await task2()  # Sets request_id="xyz" in its own copy

# Each task maintains its own isolated metadata ✅
```

### Context Vars Best Practice

This is the standard pattern for `contextvars`: treat stored values as **immutable**. Always copy-modify-set rather than mutating directly. This ensures the context isolation that `contextvars` is designed to provide.

cloud_pipelines_backend/instrumentation/contextual_logging.py
morgan-wowk: There is now 😄 

Added

```python
def delete_context_metadata(key: str) -> None:
    """Delete a metadata key from the current context.

    Similar to dict.pop() but doesn't return a value. If the key doesn't exist,
    this is a no-op (no error is raised).

    Args:
        key: The metadata key to delete (e.g., 'execution_id', 'request_id')
    """
```

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Good tip. Thank you! Done

cloud_pipelines_backend/instrumentation/api_tracing.py
morgan-wowk: Done!

cloud_pipelines_backend/instrumentation/contextual_logging.py
morgan-wowk: Great tip. Thanks!

start_local.py
morgan-wowk: Changed and noted for future! Thanks

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Good catch. Fixed. Thanks

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Done

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -137,42 +140,46 @@ def internal_process_running_executions_queue(self, session: orm.Session):
...
-                        session=session, execution=execution_node
+                except Exception as ex:
+                    _logger.exception("Error processing running container execution")
+                    session.rollback()
+                    running_container_execution.status = (
+                        bts.ContainerExecutionStatus.SYSTEM_ERROR
                     )
-                session.commit()
-            finally:
-                duration_ms = int((time.monotonic_ns() - start_timestamp) / 1_000_000)
```
Ark-kun, line 172: Hmm. Are you sure we no longer need to log processing times?

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -137,42 +140,46 @@ def internal_process_running_executions_queue(self, session: orm.Session):
...
-                for execution_node in execution_nodes:
-                    execution_node.container_execution_status = (
-                        bts.ContainerExecutionStatus.SYSTEM_ERROR
-                    )
-                    record_system_error_exception(
-                        execution=execution_node, exception=ex
+
+            # Set execution context for logging (includes container_execution_id)
+            # Get first execution_node_id for context (there may be multiple nodes using same container)
+            execution_nodes = running_container_execution.execution_nodes
```
Ark-kun: Note that you're slightly changing the logic here by making SQL queries outside the try/except context. If an exception occurs, it will go up the stack.

You might want to swap `try` and `with` contexts.

Also, AFAIK, the code now always queries the ExecutionNodes DB table whereas before it would only do that when needed. This is likely not affecting perf too much but I'd like you to understand the changes you're making (basically, you add and extra DB query to extract more information from the DB for the purpose of logging it.).

morgan-wowk: I will review this area again. My editor during development of this portion was not behaving very well, telling me the code was wrong (showing warnings) and I remember it was difficult for me satisfy. Maybe, this lead to this difference in behaviour.

morgan-wowk: > Also, AFAIK, the code now always queries the ExecutionNodes DB table whereas before it would only do that when needed. This is likely not affecting perf too much but I'd like you to understand the changes you're making (basically, you add and extra DB query to extract more information from the DB for the purpose of logging it.).

Thank you. I do understand this.

I'm curious, would you prefer to take any other approach such as passing metadata around differently to avoid a DB call? Open to any better ideas.

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: I restored duration. It should have been removed yet. Maybe in the future when we have a replacement solution

morgan-wowk: Thanks for pointing this out

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Thanks for pairing. I have implemented the changes we discussed:

* Singular `with` block
    ```python
            with contextual_logging.logging_context(
                container_execution_id=running_container_execution.id,
                execution_node_ids=execution_node_ids,
            ):
    ```
* Moving the execution_node_ids select we wrote above / outside the try catch

Also:

* Restored duration tracking

tests/test_instrumentation_request_middleware.py
Ark-kun, line 1: We should test the actual API Server (in another PR).

Issue #45, opened 2025-12-09
Run list - Search by annotations


Ark-kun:
Need a simple structural query language (expressable in JSON).

PipelineRunsService.list_annotations: https://github.com/TangleML/tangle/blob/26456ad3d56190cfec6d68d3517c135759c8d735/cloud_pipelines_backend/api_server_sql.py#L284

Predicates:

* key1 [not] exists
* key1 value [not] equals value1
* key1 value is [not] in set: value1, value2
* key1 value [not] contains substr1
* and

PR #68, opened 2026-01-13
feat: Implemented search annotations for runs

yuechao-qin:
# TODO
- [ ] Discuss if indexing and migrations is needed for annotations (key, value).
- [ ] How do I confirm that the unit test will be part of CI/CD?
- [ ] Is `EXISTS` needed for searching for Keys?  Is `EQUALS` sufficient?
- [ ] Should we combine user search in the new search API too?
- [ ] Black formatter, should it align with Google's Python style guide [here](https://google.github.io/styleguide/pyguide.html)?

# Description
Closes https://github.com/TangleML/tangle/issues/45

Implemented a new API to search annotations for runs.

## Background
Annotations are `key-value` pairs.  The following examples of annotations (e.g. `key = value`):
- `env` = `production`
- `team` = `backend`

This PR allows searches for key/value strings.

# Features
- [x] Created a new API (POST `/api/pipeline_runs/search/`) to search key/value in annotations.
- [x] Keys searchable operations
  - **EXISTS**: If any key exists regardless of key string
  - **CONTAINS**: If key contains a substring
  - **IN_SET**: If key string matches set of strings
  - **EQUALS**: If key string equals string
- [x] Values searchable operations
  - **CONTAINS**: If value contains a substring
  - **IN_SET**: If value matches set of strings
  - **EQUALS**: If value equals string
- [x] Keys and Values search operations can be negated (i.e. NOT <bool>).
- [x] N searches can be grouped (AND, OR) together and recursive.
  - Example: `(S1 and S2) or (S3 or S4)`

### 1. Key equals a string
Find runs where annotation key equals "environment":
```json
{
  "annotation_filters": {
    "filters": [
      {"operator": "equals", "key": "environment"}
    ]
  }
}
```

### 2. Key contains substring AND value in set
Find runs where key contains "env" AND value is "prod" or "staging":
```json
{
  "annotation_filters": {
    "filters": [
      {"operator": "contains", "key": "env"},
      {"operator": "in_set", "values": ["prod", "staging"]}
    ],
    "operator": "and"
  }
}
```

### 3. Complex: (key contains OR value contains) AND key NOT contains
Find runs where (key contains "env" OR any value contains "prod") AND key NOT contains "deprecated":
```json
{
  "annotation_filters": {
    "filters": [
      {
        "filters": [
          {"operator": "contains", "key": "env"},
          {"operator": "contains", "value": "prod"}
        ],
        "operator": "or"
      },
      {"operator": "contains", "value": "deprecated", "negate": true}
    ],
    "operator": "and"
  }
}
```

# Test Plan
- [x] Unit test
  - `uv run pytest tests/test_pipeline_run_search.py -v`
- [x] Manual Testing
  - Watch demo videos below
  - Add annotations for testing (PUT `/api/pipeline_runs/<ID>/annotations/<KEY>/`)
  - Query annotations for ID (GET `/api/pipeline_runs/<ID>/annotations/`)
  - Test with new search (POST `/api/pipeline_runs/search/`)
- [ ] Test on Staging (same procedure as Manual Testing above)

# Demo
https://github.com/user-attachments/assets/012008d1-3384-411d-842b-99245f3b8597

https://github.com/user-attachments/assets/0e13525c-cb1b-4990-a517-d9490b7ac9d5

### DIFF
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -310,119 +310,88 @@ class PipelineRunsApiService_Sql:
             next_page_token=next_page_token,
         )
 
-    def _build_annotation_filter_clause(
-        self,
-        *,
-        filter: AnnotationFilterType,
-    ) -> sql.ColumnElement[bool]:
-        """Build a SQLAlchemy clause from a single annotation filter."""
-        if isinstance(filter, KeyFilter):
-            return self._build_key_filter_clause(filter=filter)
-        elif isinstance(filter, ValueFilter):
-            return self._build_value_filter_clause(filter=filter)
-        elif isinstance(filter, FilterGroup):
-            return self._build_filter_group_clause(group=filter)
-        else:
-            raise ValueError(f"Unknown filter type: {type(filter)}")
-
-    def _build_key_filter_clause(
+    def _build_key_filter_condition(
         self,
         *,
         filter: KeyFilter,
     ) -> sql.ColumnElement[bool]:
-        """Build a SQLAlchemy clause for a KeyFilter.
+        """Build a SQLAlchemy condition for a KeyFilter.
 
-        KeyFilter checks for the existence or pattern of annotation keys.
+        Returns just the WHERE condition (not wrapped in EXISTS).
+        The condition is applied to the same annotation row when combined
+        with other filters in a FilterGroup.
 
-        Design Decision - EXISTS vs JOIN:
-            We use EXISTS subqueries instead of JOINs because:
-            1. Cleaner for complex boolean logic (AND/OR/NOT groups)
-            2. Automatically handles duplicates (no need for DISTINCT)
-            3. Easier to negate (~clause for NOT EXISTS)
+        Args:
+            filter: KeyFilter with operator, key/keys, and optional negate flag.
 
-            The query optimizer will use the (key, value) index on
-            PipelineRunAnnotation for efficient lookups regardless of
-            whether EXISTS or JOIN is used.
+        Returns:
+            A SQLAlchemy condition expression for the key filter.
         """
-        # Base subquery to check annotation keys
-        subquery = sql.select(bts.PipelineRunAnnotation).where(
-            bts.PipelineRunAnnotation.pipeline_run_id == bts.PipelineRun.id
-        )
-
         if filter.operator == KeyFilterOperator.EXISTS:
             # Key exists (regardless of value)
-            subquery = subquery.where(bts.PipelineRunAnnotation.key != None)
+            condition = bts.PipelineRunAnnotation.key != None
         elif filter.operator == KeyFilterOperator.EQUALS:
             # Key equals exact string
             if not filter.key:
                 raise ValueError("EQUALS operator requires 'key' to be set")
-            subquery = subquery.where(bts.PipelineRunAnnotation.key == filter.key)
+            condition = bts.PipelineRunAnnotation.key == filter.key
         elif filter.operator == KeyFilterOperator.CONTAINS:
             # Key contains substring
             if not filter.key:
                 raise ValueError("CONTAINS operator requires 'key' to be set")
-            subquery = subquery.where(
-                bts.PipelineRunAnnotation.key.contains(filter.key)
-            )
+            condition = bts.PipelineRunAnnotation.key.contains(filter.key)
         elif filter.operator == KeyFilterOperator.IN_SET:
             # Key is in a set of values
             if not filter.keys:
                 raise ValueError("IN_SET operator requires 'keys' to be set")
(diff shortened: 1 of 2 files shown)


cloud_pipelines_backend/api_router.py
```diff
@@ -269,6 +269,97 @@ def stream_container_log(
     router.get("/api/pipeline_runs/", tags=["pipelineRuns"], **default_config)(
         inject_session_dependency(list_pipeline_runs_func)
     )
+    router.post(
+        "/api/pipeline_runs/search/",
```
morgan-wowk, line 273: This is a piece of feedback on REST semantics mainly. Typically you wouldn't see a `POST` request for searching for resources, in other words getting resources. If it were a GraphQL API then things would be different.

Here's the suggestion:

Turn this into a GET endpoint:

```
GET /api/pipeline_runs
```

This is clear that you get retrieving a list of pipeline runs, which would be an unfiltered, paginated response by default. Then add the search capability after the foundation (unfiltered search) is established.

From loading the UI, I can see there is already a request being made:

```
curl 'http://localhost:8000/api/pipeline_runs?include_pipeline_names=true&include_execution_stats=true'
```

and upon asking cursor, I know that it has a pagination implementation already. We can add the filter options to this existing endpoint rather than creating a new endpoint. Which would set a good precedence for all future search capabilities on other resources.

Here is an example of how I've seen this implemented:

```
curl 'http://localhost:8000/api/pipeline_runs?filter=in(annotations.env:staging,prod)+eq(annotations.app:agenticsearch)
```

With other operators like gte (greater than or equal), lte, etc.

Accounting for complex AND / OR combinations is something that would require some thought but I wouldn't go there unless that's something we see people using. The above example is a basic OR and AND.

Ark-kun: The main reason for using POST is that GET requests do not support body. Sending complex query structures via query parameters can be problematic in certain cases. If we can overcome that, then we can use GET piepline_runs.

cloud_pipelines_backend/backend_types_sql.py
```diff
@@ -476,3 +482,19 @@ class PipelineRunAnnotation(_TableBase):
...
+
+    __table_args__ = (
+        # Index for searching pipeline runs by annotation key/value
+        # Enables efficient queries like "find runs where key='environment' and value='production'"
+        sql.Index(
+            PIPELINE_RUN_ANNOTATION_KEY_VALUE_INDEX_NAME,
+            "key",
+            "value",
+        ),
+        # Index for searching pipeline runs by annotation value only (across all keys)
```
Ark-kun, line 496: I'm not sure such feature would be useful.

cloud_pipelines_backend/database_ops.py
```diff
@@ -77,3 +77,13 @@ def migrate_db(db_engine: sqlalchemy.Engine):
...
+
+    # TODO: I believe we should create an index for the Annotation key 
+    # column and migrate existing data?
+    #
+    # Migration for annotation filtering feature
+    for index in bts.PipelineRunAnnotation.__table__.indexes:
+        if index.name == bts.PIPELINE_RUN_ANNOTATION_KEY_VALUE_INDEX_NAME:
+            index.create(db_engine, checkfirst=True)
+        if index.name == bts.PIPELINE_RUN_ANNOTATION_VALUE_INDEX_NAME:
+            index.create(db_engine, checkfirst=True)
```
Ark-kun: Let's use Black formatter.

Ark-kun: `if index.name in (A, B):`

Ark-kun:
Another question: Would we be able to reuse the same search classes and functions for component search?

cloud_pipelines_backend/backend_types_sql.py
yuechao-qin: I'm fine removing this.  Reason why I had this is explained in my other longer comment.

cloud_pipelines_backend/api_router.py
morgan-wowk: I will dig up some documentation on how I've seen this achieved previously then we can review if it meets our needs

morgan-wowk: Here are some helpful resources:

* [Google AIPs](https://google.aip.dev/160)
    * [Authorized Buyers API](https://developers.google.com/authorized-buyers/apis/guides/get-started/list-filters) 
    * [A rust implementation](https://flinect.com/blog/rust-implement-google-api-filter)
* [Bugsnag / Smartbear](https://developer.smartbear.com/bugsnag/docs/data-access-filtering)
* [Bold Commerce](https://developer.boldcommerce.com/guides/subscriptions-v2/resources/query-pagination)

morgan-wowk: I think if we solve this once, we can use the same solution for all future APIs and won't need an extra `/search` endpoint to maintain

PR #69, opened 2026-01-20
feat: Secrets

Ark-kun:
## API changes

List secrets: GET `/api/secrets/`.  
Create a secret: POST `/api/secrets/`.  
Update a secret: PUT `/api/secrets/{secret_name}`.  
Delete a secret: DELETE `/api/secrets/{secret_name}`.

## DB changes

```python
class Secret(_TableBase):
    __tablename__ = "secret"
    user_id: orm.Mapped[str] = orm.mapped_column(primary_key=True, index=True)
    secret_name: orm.Mapped[str] = orm.mapped_column(primary_key=True)
    secret_value: orm.Mapped[str]
    created_at: orm.Mapped[datetime.datetime | None] = orm.mapped_column(default=None)
    updated_at: orm.Mapped[datetime.datetime | None] = orm.mapped_column(default=None)
    extra_data: orm.Mapped[dict[str, Any] | None] = orm.mapped_column(default=None)
```

## TaskSpec changes

```yaml
componentRef: ...
  spec:
    inputs:
    - {name: input4, type: Secret}
arguments:
  input1: "constant value"
  input2:
    taskOutput:
      taskId: "task1"
      outputName: "output1"
  input3:
    graphInput:
      inputName: "graph_input_1"
  # New:
  input4:
    dynamicData:
      secret:
        name: "API_TOKEN"
```

Fixes: https://github.com/TangleML/tangle/issues/51

### DIFF
--- a/cloud_pipelines_backend/api_router.py
+++ b/cloud_pipelines_backend/api_router.py
@@ -107,6 +107,15 @@ def _setup_routes_internal(
             content={"message": str(exc)},
         )
 
+    @app.exception_handler(errors.ItemAlreadyExistsError)
+    def handle_item_already_exists_error(
+        request: fastapi.Request, exc: errors.ItemAlreadyExistsError
+    ):
+        return fastapi.responses.JSONResponse(
+            status_code=409,
+            content={"message": str(exc)},
+        )
+
     get_user_details_dependency = fastapi.Depends(user_details_getter)
 
     def get_user_name(


cloud_pipelines_backend/api_server_sql.py
```diff
@@ -994,6 +994,116 @@ def get_signed_artifact_url(
...
+        user_id: str,
+        secret_id: str,
+        secret_value: str,
+        raise_if_not_exists: bool = False,
+        raise_if_exists: bool = False,
+    ):
+        current_time = _get_current_time()
+        secret = session.get(bts.Secret, (user_id, secret_id))
+        if secret:
+            if not raise_if_exists:
```
yuechao-qin:
```suggestion
            if raise_if_exists:
```

I believe you don't want the `not` here?

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -436,6 +436,34 @@ def generate_execution_log_uri(
...
+            secret_value = secret.secret_value
+            input_artifact_data[input_name] = bts.ArtifactData(
+                total_size=len(secret_value.encode("utf-8")),
+                is_dir=False,
+                value=secret_value,
+                uri=None,
+                # This hash is not used, so we're using a dummy value here that makes it possible to identify the secret arguments in the following code.
+                hash=secret_hash,
+            )
+        session.rollback()
```
yuechao-qin, line 473: Curious why is there a rollback here?  I didn't notice any session changes.

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -994,6 +994,116 @@ def get_signed_artifact_url(
...
+            secret.updated_at = current_time
+        else:
+            if raise_if_not_exists:
+                raise errors.ItemNotFoundError(
+                    f"Secret with id '{secret_id}' does not exist."
+                )
+            secret = bts.Secret(
+                user_id=user_id,
+                secret_id=secret_id,
+                secret_value=secret_value,
```
yuechao-qin: Curious is there any validation needed for secret id/value?  For example, what if they were empty strings?

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -994,6 +994,116 @@ def get_signed_artifact_url(
...
+            session.add(secret)
+        session.commit()
+
+    def delete_secret(
+        self,
+        *,
+        session: orm.Session,
+        user_id: str,
+        secret_id: str,
+    ):
```
yuechao-qin:
```suggestion
    ) -> None:
```

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -994,6 +994,116 @@ def get_signed_artifact_url(
...
+    def _set_secret_value(
+        self,
+        *,
+        session: orm.Session,
+        user_id: str,
+        secret_id: str,
+        secret_value: str,
+        raise_if_not_exists: bool = False,
+        raise_if_exists: bool = False,
+    ):
```
yuechao-qin:
```suggestion
    ) -> None:
```

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -994,6 +994,116 @@ def get_signed_artifact_url(
...
+
+    def create_secret(
+        self,
+        *,
+        session: orm.Session,
+        user_id: str,
+        secret_id: str,
+        secret_value: str,
+    ):
+        return self._set_secret_value(
```
yuechao-qin:
```suggestion
    ) -> None:
        self._set_secret_value(
```
`_set_secret_value` doesn't return anything, so a `return` is not necessary here?

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -994,6 +994,116 @@ def get_signed_artifact_url(
...
+
+    def update_secret(
+        self,
+        *,
+        session: orm.Session,
+        user_id: str,
+        secret_id: str,
+        secret_value: str,
+    ):
+        return self._set_secret_value(
```
yuechao-qin:
```suggestion
    ) -> None:
        self._set_secret_value(
```

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -994,6 +994,116 @@ def get_signed_artifact_url(
         return GetArtifactSignedUrlResponse(signed_url=signed_url)
 
 
+# === Secrets Service
+@dataclasses.dataclass(kw_only=True)
+class SecretInfoResponse:
+    secret_id: str
```
maxy-shpfy: @Ark-kun we may need to provide more data to display on UI - e.g. createdDate or description or expirationDate

cloud_pipelines_backend/api_router.py
```diff
@@ -390,6 +390,30 @@ def get_current_user(
...
+        inject_session_dependency(
+            inject_user_name(secrets_service.list_secrets, parameter_name="user_id")
+        )
+    )
+    router.post("/api/secrets/", tags=["secrets"], **default_config)(
+        inject_session_dependency(
+            inject_user_name(secrets_service.create_secret, parameter_name="user_id")
+        )
+    )
+    router.put("/api/secrets/{secret_id}", tags=["secrets"], **default_config)(
```
maxy-shpfy: I see that `secret_value` is GET (Query) parameter. It is not secured since Query params may be traced via logs etc. It should be part of the body

maxy-shpfy: Same applies to all other APIs

cloud_pipelines_backend/component_structures.py
```diff
@@ -317,7 +317,23 @@ class TaskOutputArgument(_BaseModel):  # Has additional constructor for convenie
     task_output: TaskOutputReference
 
 
-ArgumentType = Union[PrimitiveTypes, GraphInputArgument, TaskOutputArgument]
+@dataclasses.dataclass
+class SecretReference(_BaseModel):
+    """References a secret"""
+
+    id: str
```
maxy-shpfy: In the Tech design we're using `name` not `id`. What should be used?

cloud_pipelines_backend/api_server_sql.py
Ark-kun: Good catch. Fixed.

cloud_pipelines_backend/orchestrator_sql.py
Ark-kun: It's likely not needed here, but it's a pattern I'm using in some places.
The reason is that SqlAlchemy starts transactions automatically. Transactions that read and write common rows can conflict. For example situations like this:
* TransactionA reads key1=value1
* TransactionB writes key1=value2
* TransactionB is committed
* TransactionA writes key2=value1
* TransactionA is committed

I'm trying to avoid such conflicts. It might be possible to solve by explicitly specifying isolation labels, but it can also be solved by making smaller, more granular transactions.
Transaction can be ended by either commit or rollback. When I'm not making any changes, I use rollback to signify that.

cloud_pipelines_backend/api_server_sql.py
Ark-kun: Done.

cloud_pipelines_backend/api_server_sql.py
Ark-kun: I made it return the secret info.

cloud_pipelines_backend/component_structures.py
Ark-kun: I've renamed secret "id" to "name" everywhere based on your feedback.

cloud_pipelines_backend/api_server_sql.py
Ark-kun: These can always be added later if needed. However I've added the available info - `created_at` and `updated_at`

cloud_pipelines_backend/api_server_sql.py
Ark-kun: Good idea. Added validation (empty secret name).

cloud_pipelines_backend/api_server_sql.py
Ark-kun: Done.

cloud_pipelines_backend/api_router.py
Ark-kun: Great catch! I've moved secret_value to HTTP request body.

cloud_pipelines_backend/api_server_sql.py
Ark-kun: Done

cloud_pipelines_backend/backend_types_sql.py
```diff
@@ -476,3 +477,13 @@ class PipelineRunAnnotation(_TableBase):
...
     value: orm.Mapped[str | None] = orm.mapped_column(default=None)
+
+
+class Secret(_TableBase):
+    __tablename__ = "secret"
+    user_id: orm.Mapped[str] = orm.mapped_column(primary_key=True, index=True)
+    secret_name: orm.Mapped[str] = orm.mapped_column(primary_key=True)
+    secret_value: orm.Mapped[str]
+    created_at: orm.Mapped[datetime.datetime | None] = orm.mapped_column(default=None)
+    updated_at: orm.Mapped[datetime.datetime | None] = orm.mapped_column(default=None)
```
yuechao-qin: When would created and updated datetime ever be none?

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -1102,6 +1228,23 @@ def _recursively_create_all_executions_and_artifacts(
...
+        if isinstance(input_artifact_node, structures.DynamicDataArgument):
+            # We don't use these secret arguments, but adding them just in case.
+            extra_data = root_execution_node.extra_data or {}
+            dynamic_data_arguments = extra_data.setdefault(
+                bts.EXECUTION_NODE_EXTRA_DATA_DYNAMIC_DATA_ARGUMENTS_KEY, {}
+            )
+            dynamic_data_arguments[input_spec.name] = input_artifact_node.dynamic_data
+            if not (
+                isinstance(input_artifact_node.dynamic_data, str)
+                or len(input_artifact_node.dynamic_data) == 1
```
yuechao-qin:
```suggestion
                isinstance(input_artifact_node.dynamic_data, str)
                or (isinstance(input_artifact_node.dynamic_data, dict) and len(input_artifact_node.dynamic_data) == 1)
```
Would it make sense to be more explicit here since a (list, set, bytes) can pass the logic `len(x) == 1`?

Also, would it make sense to do this logic check first in this code block?

cloud_pipelines_backend/launchers/interfaces.py
```diff
@@ -36,6 +36,7 @@ class InputArgument:
     value: str | None = None
     uri: str | None = None
     staging_uri: str
+    is_secret: bool = False
```
yuechao-qin, line 39: If I'm understanding your design, would it make sense to make this a `DynamicType | None`, which DynamicType is a StrEum that only has `SECRET` for now, so that it can scale with other Dynamic Type supports in the future?

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -458,6 +439,39 @@ def generate_execution_log_uri(
...
+                secret_value = secret.secret_value
+                input_artifact_data[input_name] = bts.ArtifactData(
+                    total_size=len(secret_value.encode("utf-8")),
+                    is_dir=False,
+                    value=secret_value,
+                    uri=None,
+                    # This hash is not used, so we're using a dummy value here that makes it possible to identify the secret arguments in the following code.
+                    hash=secret_hash,
+                )
+        session.rollback()
```
yuechao-qin, line 473: Can you add a comment here then to signify your intent of the rollback.  

Another suggestion would be to have functions that only has a commit/rollback that the function name/docstring is clear on the intent of the use.

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -998,6 +994,129 @@ def get_signed_artifact_url(
...
+        secret_name: str,
+        secret_value: str,
+        raise_if_not_exists: bool = False,
+        raise_if_exists: bool = False,
+    ) -> SecretInfoResponse:
+        current_time = _get_current_time()
+        secret = session.get(bts.Secret, (user_id, secret_name))
+        if secret:
+            if raise_if_exists:
+                raise errors.ItemAlreadyExistsError(
```
yuechao-qin, line 1083: I didn't notice an exception handler for `ItemAlreadyExistsError` in `api_router.py`?  I see `ItemNotFoundError` and `PermissionError` exists.

```
    @app.exception_handler(errors.ItemAlreadyExistsError)
    def handle_already_exists_error(request: fastapi.Request, exc: errors.ItemAlreadyExistsError):
        return fastapi.responses.JSONResponse(
            status_code=409,
            content={"message": str(exc)},
        )
```

cloud_pipelines_backend/launchers/interfaces.py
Ark-kun: Good point. However I think secret is special here. Let me explain:
This structure describes actual artifacts with known data (that have size and value or URI).
Some dynamicData arguments do not have actual data since they are only resolved by the launcher.
But arguments like secrets do have data, so can use this structure.
However, we want the launcher to try to hide some values (e.g. secrets). This makes these secret values special. This is not relevant for other argument types, just secrets.

Maybe I should just remove this value since it's not used yet.

Ark-kun: Removed for now.

cloud_pipelines_backend/backend_types_sql.py
Ark-kun: Made them required.

cloud_pipelines_backend/orchestrator_sql.py
Ark-kun: Added comment.

cloud_pipelines_backend/api_server_sql.py
Ark-kun: Done.

cloud_pipelines_backend/api_server_sql.py
Ark-kun: Done.

PR #76, opened 2026-02-02
feat: Add pipeline run outcome metrics

morgan-wowk:
# Pipeline Metrics Instrumentation

Tracks pipeline lifecycle from creation to completion with labels for status and user.

Metrics added:
- pipeline_runs_total: Counter tracking pipeline runs by status (running/succeeded/failed/cancelled) and created_by
- pipeline_run_duration_seconds: Histogram tracking total pipeline duration by final status

These metrics provide visibility into pipeline success rates, completion times, and usage patterns per user.
Durations measure total lifecycle time from creation to terminal state (including queue and execution time).

### DIFF
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -30,6 +30,7 @@ from . import component_structures as structures
 from . import backend_types_sql as bts
 from . import errors
 from .errors import ItemNotFoundError
+from .instrumentation import metrics
 
 
 # ==== PipelineJobService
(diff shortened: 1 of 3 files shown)


Ark-kun:
I'm not fully sure we should be doing this on the app side.
Backend can be restarted at any time (e.g. new version is deployed). With current implementation in the PR, this seems to affect the metrics. But I think it shouldn't.

I'm not sure about reporting pipeline run times like this. Tracking execution durations might be more OK.

I'm not sure the backend should be aggregating run's execution status statistics into a single status. This single status will likely be not useful for many of the users. I think it's better when the users can get the full status stats and UI can provide derivatives.

Ark-kun:
Overall idea: We want to leverage the telemetry processing systems and dashboards as much as possible to avoid re-inventing the wheel.

Ideal case: We report events via OTel and the dedicated metrics system (which we do not re-invent) creates metrics from those events.
Realistic compromise case: In addition to events, we report some simple metrics that we get anyways (or are cheap to get). E.g. time to process a queued/running execution.
Case we want to avoid (at least for now): Calculating and tracking complex metrics ourselves on the backend side.

PR #92, opened 2026-02-13
fix: Contextual logging import to use absolute path

morgan-wowk:
Changed the import in structured_logging.py from relative to absolute to ensure that the same contextual_logging module instance is used across all imports. This fixes an issue where context metadata set in one module was not visible in another due to separate module instances having separate ContextVar objects.

This ensures LoggingContextFilter and contextual_logging.logging_context share the same _context_metadata ContextVar, allowing contextual data like container_execution_id and execution_node_ids to appear in logs.

### DIFF
--- a/cloud_pipelines_backend/instrumentation/structured_logging.py
+++ b/cloud_pipelines_backend/instrumentation/structured_logging.py
@@ -6,7 +6,7 @@ contextual_logging module to automatically include context metadata in log recor
 
 import logging
 
-from . import contextual_logging
+from cloud_pipelines_backend.instrumentation import contextual_logging
 
 
 class LoggingContextFilter(logging.Filter):


cloud_pipelines_backend/instrumentation/structured_logging.py
```diff
@@ -6,7 +6,7 @@
 
 import logging
 
-from . import contextual_logging
+from cloud_pipelines_backend.instrumentation import contextual_logging
```
morgan-wowk, line 9: We can look into preventing this type of import in the future with a PR check

Ark-kun:
Hmm. The module should still be imported only once. If relative imports always caused such issues, then `isinstance` checks won't have worked.

Ark-kun:
Fixed by https://github.com/Shopify/oasis-backend/pull/107

PR #94, opened 2026-02-17
feat: Model to track node states and has_ended flag

yuechao-qin:
### TL;DR

Added an `ExecutionSummary` class to track execution state with comprehensive test coverage.

### What changed?

- Created a new `ExecutionSummary` dataclass in `api_server_sql.py` that tracks:
  - Total number of nodes
  - Number of ended nodes
  - Whether the entire execution has ended
- Added a `count_node_status` method to update counters based on node status
- Created a new test file `test_api_server_sql.py`.

### How to test?

Run the new test suite:
```
```

### Why make this change?

This will be used in upstream PRs to calculate execution summary for runs and root nodes.

### DIFF
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -475,6 +475,22 @@ class ArtifactNodeIdResponse:
     id: bts.IdType
 
 
+@dataclasses.dataclass(kw_only=True)
+class ExecutionStatusSummary:
+    total_executions: int = 0
+    ended_executions: int = 0
+    has_ended: bool = False
+
+    def count_execution_status(
+        self, *, status: bts.ContainerExecutionStatus, count: int
+    ) -> None:
+        self.total_executions += count
+        if status in bts.CONTAINER_STATUSES_ENDED:
+            self.ended_executions += count
+
+        self.has_ended = self.ended_executions == self.total_executions
+
+
 @dataclasses.dataclass
 class GetGraphExecutionStateResponse:
     child_execution_status_stats: dict[bts.IdType, dict[str, int]]
--- a/tests/test_api_server_sql.py
+++ b/tests/test_api_server_sql.py
@@ -1,0 +1,57 @@
+from cloud_pipelines_backend import backend_types_sql as bts
+from cloud_pipelines_backend.api_server_sql import ExecutionStatusSummary
+
+
+class TestExecutionStatusSummary:
+    def test_initial_state(self):
+        summary = ExecutionStatusSummary()
+        assert summary.total_executions == 0
+        assert summary.ended_executions == 0
+        assert summary.has_ended is False
+
+    def test_accumulate_all_ended_statuses(self):
+        """Add each ended status with 2^i count for robust uniqueness."""
+        summary = ExecutionStatusSummary()
+        ended_statuses = sorted(bts.CONTAINER_STATUSES_ENDED, key=lambda s: s.value)
+        expected_total = 0
+        expected_ended = 0
+        for i, status in enumerate(ended_statuses):
+            count = 2**i
+            summary.count_execution_status(status=status, count=count)
+            expected_total += count
+            expected_ended += count
+            assert summary.total_executions == expected_total
+            assert summary.ended_executions == expected_ended
+            assert summary.has_ended is True
+
+    def test_accumulate_all_in_progress_statuses(self):
+        """Add each in-progress status with 2^i count for robust uniqueness."""
+        summary = ExecutionStatusSummary()
+        in_progress_statuses = sorted(
+            set(bts.ContainerExecutionStatus) - bts.CONTAINER_STATUSES_ENDED,
+            key=lambda s: s.value,
+        )
+        expected_total = 0
+        for i, status in enumerate(in_progress_statuses):
+            count = 2**i
+            summary.count_execution_status(status=status, count=count)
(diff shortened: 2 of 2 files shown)


cloud_pipelines_backend/api_server_sql.py
```diff
@@ -475,6 +475,22 @@ class ArtifactNodeIdResponse:
     id: bts.IdType
 
 
+@dataclasses.dataclass(kw_only=True)
+class ExecutionSummary:
+    total_nodes: int = 0
+    ended_nodes: int = 0
+    has_ended: bool = False
+
+    def count_node_status(
```
Ark-kun: Maybe, instead of adding a single status info, we can have a class method to convert a whole status stats dict to an ExecutionStatusSummary instance?

tests/test_api_server_sql.py
```diff
@@ -0,0 +1,50 @@
...
+        for i, status in enumerate(in_progress_statuses, start=1):
+            summary.count_node_status(status=status, count=1)
+            assert summary.total_nodes == i
+            assert summary.ended_nodes == 0
+            assert summary.has_ended is False
+
+    def test_accumulate_all_statuses(self):
+        """Add every ContainerExecutionStatus one by one, confirming counts after each."""
+        summary = ExecutionSummary()
+        all_statuses = sorted(bts.ContainerExecutionStatus, key=lambda s: s.value)
```
Ark-kun, line 46: We can have a test where we calculate summary numbers for concrete status stats dict. If each i-th status has 2^i count, then the summary counts will be pretty robust.

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -475,6 +475,22 @@ class ArtifactNodeIdResponse:
...
+        self, *, status: bts.ContainerExecutionStatus, count: int
+    ) -> None:
+        self.total_nodes += count
+        if status in bts.CONTAINER_STATUSES_ENDED:
+            self.ended_nodes += count
+
+        self.has_ended = self.total_nodes > 0 and self.ended_nodes == self.total_nodes
```
Ark-kun: I'd say 0 out of 0 is "ended" too.

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -475,6 +475,22 @@ class ArtifactNodeIdResponse:
     id: bts.IdType
 
 
+@dataclasses.dataclass(kw_only=True)
+class ExecutionSummary:
```
Ark-kun: `Execution**Status**Summary`

cloud_pipelines_backend/api_server_sql.py
yuechao-qin: I was thinking the same thing, but when I was using this dataclass in the upstream PRs, the API responses are being created through a loop with one status at a time, so for performance reasons, this function fits into the existing loop well.

Let me know if you have any concerns leaving this the way it is.

Ark-kun: I've looked at the usage, but I'm not sure there is a problem with just passing a status_stats dict here.

```python
class ExecutionStatusSummary:
    def from_status_stats(status_stats: dict[bts.ContainerExecutionStatus, int]) -> ExecutionStatusSummary:
        for status, count in status_stats.items():
            ...
```

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -475,6 +475,22 @@ class ArtifactNodeIdResponse:
     id: bts.IdType
 
 
+@dataclasses.dataclass(kw_only=True)
+class ExecutionStatusSummary:
+    total_nodes: int = 0
```
Ark-kun: `total_executions`, `ended_executions`?

cloud_pipelines_backend/api_server_sql.py
yuechao-qin: @Ark-kun maybe I'm having trouble seeing your perspective, but there are two APIs that use the `ExecutionStatusSummary`:

- `GET /api/pipeline_runs/`
```
  "execution_status_stats": {
    "SUCCEEDED": 4
  },
```

- `GET 
/api/executions/{id}/graph_execution_state`
```
  "child_execution_status_stats": {
    "019c78b69e0e376ffedf": {
      "SUCCEEDED": 1
    },
    "019c78b69e0e60f071db": {
      "SUCCEEDED": 1
    },
    "019c78b69e0e9b4ab0ff": {
      "SUCCEEDED": 1
    },
    "019c78b69e0ef7f5d003": {
      "SUCCEEDED": 1
    }
  },
```

So, having a running function "count" helps with both APIs because:
- `pipeline_runs` does work better with your approach since it'll result in a single dict of `states <-> count`
- `graph_execution_state` I believe doesn't work well because there'll be multiple children dicts and the dicts are still accumulating in the loop.


- `pipeline_runs`

- `graph_execution_state`


If you're saying to iterate over the final responses (single dict, multiple children dicts), I'm fine doing that.  It won't be a class method, it'll be an instance method, to support multiple children dicts.

Ark-kun: I was thinking that in either case we'll have a single final dict of status counts. But you're correct that for execution status API we have multiple dicts.  
  
Let's proceed.

morgan-wowk:
I know this is already merged but reviewing for my own knowledge. Looks good! :)

PR #95, opened 2026-02-17
refactor: API Server - Pipeline runs - Added a helper function to get execution status stats

yuechao-qin:
### TL;DR

Refactored execution status stats calculation and added comprehensive tests for the pipeline runs API service.

### What changed?

- Extracted the execution status stats formatting logic into a new method `_get_execution_status_stats` 
- Added extensive test coverage for the `PipelineRunsApiService_Sql` class

### How to test?

Run the new test cases in `tests/test_api_server_sql.py`:

```bash
```

### Why make this change?

Created this function to be used in upstream PRs for Pipeline Run API (list and get)

### DIFF
--- a/tests/test_api_server_sql.py
+++ b/tests/test_api_server_sql.py
@@ -1,5 +1,6 @@
 import datetime
 import json
+from collections.abc import Callable
 
 import pytest
 import sqlalchemy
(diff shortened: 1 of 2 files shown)


tests/test_api_server_sql.py
```diff
@@ -1,5 +1,50 @@
+from sqlalchemy import orm
+
 from cloud_pipelines_backend import backend_types_sql as bts
-from cloud_pipelines_backend.api_server_sql import ExecutionStatusSummary
+from cloud_pipelines_backend import database_ops
+from cloud_pipelines_backend.api_server_sql import (
+    ExecutionStatusSummary,
+    PipelineRunsApiService_Sql,
```
morgan-wowk: :nit   
  
If we are treating test code with the same standard as production code then I'll just mention:

Alexey has asked me to import modules (`api_server_sql`) rather than directly importing classes or function.

So you would use `api_server_sql.ExecutionStatusSummary` rather than `ExecutionStatusSummary` directly in this file. Same with other imports.

tests/test_api_server_sql.py
```diff
@@ -55,3 +100,59 @@ def test_accumulate_all_statuses(self):
...
             assert summary.has_ended == (expected_ended == expected_total)
+
+
+class TestPipelineRunServiceList:
+    def test_list_empty(self):
+        session_factory = _initialize_db_and_get_session_factory()
+        service = PipelineRunsApiService_Sql()
+        with session_factory() as session:
+            result = service.list(session=session)
+            assert result.pipeline_runs == []
```
morgan-wowk, line 1848: I would be interested to establish a test structure with you and Alexey in the future.  
  
From my Go programming days, we established a structure and this allowed the team to be able to easily read new tests because there were already familiar with the consistent way they were written.  
  
For now, it's more important that we practice writing tests, but would love to get to a point where they are consistently structured in the future :)

tests/test_api_server_sql.py
```diff
@@ -55,3 +100,59 @@ def test_accumulate_all_statuses(self):
...
+
+        with session_factory() as session:
+            result = service.list(session=session)
+            assert len(result.pipeline_runs) == 1
+            assert result.pipeline_runs[0].root_execution_id == root_id
+            assert result.pipeline_runs[0].created_by == "user1"
+            assert result.pipeline_runs[0].execution_status_stats is None
+
+    def test_list_with_execution_stats(self):
+        session_factory = _initialize_db_and_get_session_factory()
```
morgan-wowk, line 1868: We call `_initialize_db_and_get_session_factory` and migrate the DB for every test.   
  
For this whole file, are we leaving sessions open when we should be closing them at times they are no longer needed? For example, we run one test, open some sessions, then run another test, open a new session, but never close them as we're done with them?

morgan-wowk: Asking so that we avoid any build up of memory, resources, or sessions that will bite us in the future as we add more tests.

morgan-wowk (changes requested):
I can see that it is functionally the same from a production code point of view. Just left some comments on the tests.  
  
Thanks for doing this!

tests/test_api_server_sql.py
Ark-kun: Hmm. I think we are closing the session via exiting the context. We get "session factory", not "session" here.
The DB engine remains though.

morgan-wowk: Ah yes looks like it.

tests/test_api_server_sql.py
```diff
@@ -1,5 +1,50 @@
...
+    """Create an ExecutionToAncestorExecutionLink."""
+    link = bts.ExecutionToAncestorExecutionLink(
+        ancestor_execution=ancestor_node,
+        execution=execution_node,
+    )
+    session.add(link)
+    session.flush()
+
+
+def _create_pipeline_run(session, root_execution, created_by=None, annotations=None):
```
Ark-kun: Let's use Python type annotations everywhere in function signatures.

tests/test_api_server_sql.py
```diff
@@ -1,5 +1,50 @@
...
+    """Helper to create a PipelineRun linked to a root execution node."""
+    run = bts.PipelineRun(root_execution=root_execution)
+    if created_by:
```
Ark-kun: This conditionals are not needed. You can just construct the PipelineRun object.

Ark-kun (changes requested):

morgan-wowk (approved):

Ark-kun (approved):

Ark-kun (approved):

yuechao-qin:
### Merge activity

* **Apr 27, 7:45 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/95).
* **Apr 27, 7:46 PM UTC**: @yuechao-qin merged this pull request with [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/95).

Issue #87, opened 2026-02-09
API Server - Add `has_ended` to the response of PieplineRunsService.list


Ark-kun:
Add `has_ended: bool | None` to PipelineRunResponse which is set when `list(..., include_execution_stats == True)`

PR #96, opened 2026-02-17
feat: API Server - Pipeline runs - Added execution summary to the list method

yuechao-qin:
### TL;DR

Closes https://github.com/TangleML/tangle/issues/87

Added execution summary to pipeline run responses to provide aggregated information about pipeline execution status  (`GET /api/pipeline_runs/`)..

### What changed?

- Added `execution_summary` field to `PipelineRunResponse` dataclass
- Renamed `_get_execution_status_stats` to `_get_execution_stats_and_summary` to return both stats and summary

### How to test?

1. Create a pipeline run with multiple nodes in different states
2. Call the list API with `include_execution_stats=True`
3. Verify the response contains both execution stats and execution summary
4. Check that the summary correctly reports total nodes, ended nodes, and completion status

Run the new test cases in `tests/test_api_server_sql.py`:

```bash
pytest tests/test_api_server_sql.py
```

### Why make this change?

This closes https://github.com/TangleML/tangle/issues/87

PR #97, opened 2026-02-17
feat: API Server - Pipeline runs - Made it possible to get pipeline run execution summary when using the get method

yuechao-qin:
### TL;DR

Added execution statistics to the pipeline run get endpoint (`GET /api/pipeline_runs/{id}`).

### What changed?

- Enhanced the `get` method to optionally include execution statistics
- Added a new parameter `include_execution_stats` (default: False) to the `get` method
- Extracted the execution stats population logic into a new helper method `_populate_execution_stats`

### How to test?

1. Call the pipeline run get endpoint with `include_execution_stats=True`
2. Verify that the response includes execution status statistics and summary
3. Run the new tests in `TestPipelineRunServiceGet` class to ensure the functionality works as expected

Run the new test cases in `tests/test_api_server_sql.py`:

```bash
```

### Why make this change?

- Related to https://github.com/TangleML/tangle/issues/87
- `tangle-deploy` needs to know if a specific pipeline run is completed

### DIFF
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -157,11 +157,21 @@ class PipelineRunsApiService_Sql:
         session.refresh(pipeline_run)
         return PipelineRunResponse.from_db(pipeline_run)
 
-    def get(self, session: orm.Session, id: bts.IdType) -> PipelineRunResponse:
+    def get(
+        self,
+        session: orm.Session,
+        id: bts.IdType,
+        include_execution_stats: bool = False,
+    ) -> PipelineRunResponse:
         pipeline_run = session.get(bts.PipelineRun, id)
         if not pipeline_run:
             raise errors.ItemNotFoundError(f"Pipeline run {id} not found.")
-        return PipelineRunResponse.from_db(pipeline_run)
+        response = PipelineRunResponse.from_db(pipeline_run)
+        if include_execution_stats:
+            response = self._populate_execution_stats(
+                session=session, response=response
+            )
+        return response
 
     def terminate(
         self,
@@ -275,12 +285,22 @@ class PipelineRunsApiService_Sql:
                     )
             response.pipeline_name = pipeline_name
         if include_execution_stats:
-            stats, summary = self._get_execution_stats_and_summary(
-                session=session,
-                root_execution_id=pipeline_run.root_execution_id,
+            response = self._populate_execution_stats(
+                session=session, response=response
             )
-            response.execution_status_stats = stats
-            response.execution_summary = summary
+        return response
+
+    def _populate_execution_stats(
+        self,
+        session: orm.Session,
+        response: PipelineRunResponse,
+    ) -> PipelineRunResponse:
+        stats, summary = self._get_execution_stats_and_summary(
+            session=session,
+            root_execution_id=response.root_execution_id,
+        )
+        response.execution_status_stats = stats
+        response.execution_summary = summary
         return response
 
     def _get_execution_stats_and_summary(
--- a/tests/test_api_server_sql.py
+++ b/tests/test_api_server_sql.py
@@ -1893,3 +1893,67 @@ class TestPipelineRunServiceList:
             assert summary.total_executions == 2
             assert summary.ended_executions == 1
             assert summary.has_ended is False
+
+
+class TestPipelineRunServiceGet:
+    def test_get_not_found(self):
+        session_factory = _initialize_db_and_get_session_factory()
+        service = api_server_sql.PipelineRunsApiService_Sql()
+        with session_factory() as session:
+            with pytest.raises(errors.ItemNotFoundError):
+                service.get(session=session, id="nonexistent-id")
+
+    def test_get_returns_pipeline_run(self):
+        session_factory = _initialize_db_and_get_session_factory()
+        service = api_server_sql.PipelineRunsApiService_Sql()
+        with session_factory() as session:
+            root = _create_execution_node(session=session)
+            root_id = root.id
+            run = _create_pipeline_run(session=session, root_execution=root)
+            run_id = run.id
+            session.commit()
+
+        with session_factory() as session:
+            result = service.get(session=session, id=run_id)
+            assert result.id == run_id
+            assert result.root_execution_id == root_id
+            assert result.execution_status_stats is None
+            assert result.execution_summary is None
+
+    def test_get_with_execution_stats(self):
+        session_factory = _initialize_db_and_get_session_factory()
+        service = api_server_sql.PipelineRunsApiService_Sql()
+        with session_factory() as session:
+            root = _create_execution_node(session=session)
+            root_id = root.id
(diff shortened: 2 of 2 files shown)


yuechao-qin:
* **#99** <a href="https://app.graphite.com/github/pr/TangleML/tangle/99?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#98** <a href="https://app.graphite.com/github/pr/TangleML/tangle/98?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#97** <a href="https://app.graphite.com/github/pr/TangleML/tangle/97?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/97?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#96** <a href="https://app.graphite.com/github/pr/TangleML/tangle/96?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#95** <a href="https://app.graphite.com/github/pr/TangleML/tangle/95?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94** <a href="https://app.graphite.com/github/pr/TangleML/tangle/94?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -274,6 +277,18 @@ def create_pipeline_run_response(
             next_page_token=next_page_token,
         )
 
+    def _populate_execution_stats(
```
Ark-kun, line 293: In general, I prefer functions that return value, rather than methods that modify things in-palce.

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -130,11 +130,19 @@ def create(
         session.refresh(pipeline_run)
         return PipelineRunResponse.from_db(pipeline_run)
 
-    def get(self, session: orm.Session, id: bts.IdType) -> PipelineRunResponse:
+    def get(
+        self,
+        session: orm.Session,
+        id: bts.IdType,
+        include_execution_stats: bool = False,
```
morgan-wowk, line 164: To each their own on this:

I prefer to use an `expand` class for these use cases, which is more extensible in the future.

```
@dataclasses.dataclass(frozen=True)
class PipelineRunResponseExpandOptions:
    """A type-safe structure to define which fields to expand."""
    execution_stats: bool = False
```

then using it as:

```
if expand.execution_stats:
    response = self._populate_execution_stats(session=session, response=response)
```

morgan-wowk: You could totally say "pre-mature class" for just one expansion, but I find the consistency and readiness preferable over future refactors or inconsistency.

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -130,11 +130,19 @@ def create(
...
+    ) -> PipelineRunResponse:
         pipeline_run = session.get(bts.PipelineRun, id)
         if not pipeline_run:
             raise ItemNotFoundError(f"Pipeline run {id} not found.")
-        return PipelineRunResponse.from_db(pipeline_run)
+        response = PipelineRunResponse.from_db(pipeline_run)
+        if include_execution_stats:
```
morgan-wowk, line 170: Maybe we should expand the PipelineRun rather than the response? It provides more value to the rest of the codebase that might want to expand this information in other places later on.

This suggestion implies moving the logic to the service layer (e.g. a PipelineRunsService rather than what is currently PipelineRunsApiService_Sql) and adding the field to the PipelineRun model rather than onto the response model.

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -274,6 +277,19 @@ def create_pipeline_run_response(
             next_page_token=next_page_token,
         )
 
+    def _populate_execution_stats(
```
morgan-wowk, line 293: I would prefer to put general-purpose business logic on a separate class such as a PipelineRunsService rather than having functions like this in a module, and on a class, that is specifically tied to the http transport layer.

morgan-wowk (changes requested):

cloud_pipelines_backend/api_server_sql.py
morgan-wowk: PS: This relates to the fact that there might be opportunities to better organize our backend code in general. For example, orchestrator_sql.py has a lot going on.

cloud_pipelines_backend/api_server_sql.py
yuechao-qin: I agree, I like this pattern too, I tend to do this when there are > 5 params.

cloud_pipelines_backend/api_server_sql.py
yuechao-qin: I think that's a good idea.  Happy to do that in a separate PR, I'm hoping to keep the scope minimal for this PR.

cloud_pipelines_backend/api_server_sql.py
yuechao-qin: I agree more organization is definitely needed in this file.  I have another PR https://app.graphite.com/github/pr/TangleML/tangle/199, that's helping doing some of this refactoring.

TLDR: There is a new `/execution/*` folder that this stack of PR logic could fit nicely in.  I'm just waiting on landing that and then can do cleanup in a future PR.

cloud_pipelines_backend/api_server_sql.py
morgan-wowk: @yuechao-qin Seems like this comment is not addressed

morgan-wowk (approved):

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -274,12 +284,22 @@ def _create_pipeline_run_response(
...
-                session=session,
-                root_execution_id=pipeline_run.root_execution_id,
+            response = self._populate_execution_stats(
+                session=session, response=response
             )
-            response.execution_status_stats = stats
-            response.execution_summary = summary
+        return response
+
+    def _populate_execution_stats(
```
Ark-kun, line 293: Let's not extract this function unless we need that in 2-3+ places.

Ark-kun (approved):
Approved, but please resolve the threads.

yuechao-qin:
### Merge activity

* **Apr 27, 7:45 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/97).
* **Apr 27, 7:49 PM UTC**: [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/97) rebased this pull request as part of a merge.
* **Apr 27, 7:50 PM UTC**: @yuechao-qin merged this pull request with [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/97).

PR #107, opened 2026-02-24
feat: Search annotations in pipeline run API

yuechao-qin:
### TL;DR

Implemented annotation-based filtering for Pipeline Run API and proper pagination handling.

### What changed?

API `GET /api/pipeline_runs/`

#### Functionality

- Implemented support for `filter_query` which handles annotation searches in pipeline runs.
- Added new database index for annotation-based queries.  Covering index `(key, pipeline_run_id, value)`.
- Enhanced pagination to preserve both legacy `filter` and new `filter_query` across pages.

#### Other

- Created new `filter_query_sql.py` module with comprehensive filter query processing:
  - `PageToken` class for encoding/decoding pagination state
  - Support for annotation-based predicates: `key_exists`, `value_equals`, `value_contains`, `value_in`
  - Logical operators: `and`, `or`, `not` with proper nesting
  - Integration with existing legacy filter system
- Refactored pipeline run listing to use the new filter query system while maintaining backward compatibility

### How to test?

```
```

- Test basic annotation filtering: `{"and": [{"key_exists": {"key": "team"}}]}`
- Test value-based filtering: `{"and": [{"value_equals": {"key": "env", "value": "prod"}}]}`
- Test logical combinations with nested `and`/`or`/`not` predicates
- Verify mutual exclusivity validation between `filter` and `filter_query` parameters
- Test pagination preservation of filter_query state across multiple pages
- Confirm legacy filter functionality remains unchanged

### Why make this change?

Annotation search support is the backbone for all the upcoming search features (`create_by`, `pipeline name`, `status`) for Pipeline Runs.  The other searches will be search against the annotations table.

### DIFF
--- a/cloud_pipelines_backend/backend_types_sql.py
+++ b/cloud_pipelines_backend/backend_types_sql.py
@@ -2,7 +2,7 @@ import dataclasses
 import datetime
 import enum
 import typing
-from typing import Any
+from typing import Any, Final
 
 import sqlalchemy as sql
 from sqlalchemy import orm
@@ -294,6 +294,9 @@ class ExecutionToAncestorExecutionLink(_TableBase):
 # So we need to jump through extra hoops to make the relationship many-to-many again.
 class ExecutionNode(_TableBase):
     __tablename__ = "execution_node"
+    _IX_EXECUTION_NODE_CACHE_KEY: Final[str] = (
+        "ix_execution_node_container_execution_cache_key"
+    )
     id: orm.Mapped[IdType] = orm.mapped_column(
         primary_key=True, init=False, insert_default=generate_unique_id
     )
@@ -491,6 +494,9 @@ class ContainerExecution(_TableBase):
 
 class PipelineRunAnnotation(_TableBase):
     __tablename__ = "pipeline_run_annotation"
+    _IX_ANNOTATION_RUN_ID_KEY_VALUE: Final[str] = (
+        "ix_pipeline_run_annotation_pipeline_run_id_key_value"
+    )
     pipeline_run_id: orm.Mapped[IdType] = orm.mapped_column(
         sql.ForeignKey(PipelineRun.id),
         primary_key=True,
@@ -499,6 +505,15 @@ class PipelineRunAnnotation(_TableBase):
     pipeline_run: orm.Mapped[PipelineRun] = orm.relationship(repr=False, init=False)
     key: orm.Mapped[str] = orm.mapped_column(default=None, primary_key=True)
     value: orm.Mapped[str | None] = orm.mapped_column(default=None)
+
+    __table_args__ = (
+        sql.Index(
+            _IX_ANNOTATION_RUN_ID_KEY_VALUE,
+            "pipeline_run_id",
+            "key",
+            "value",
+        ),
+    )
 
 
 class Secret(_TableBase):
--- a/cloud_pipelines_backend/component_structures.py
+++ b/cloud_pipelines_backend/component_structures.py
@@ -54,7 +54,6 @@ import pydantic
 import pydantic.alias_generators
 from pydantic.dataclasses import dataclass as pydantic_dataclasses
 
-
 # PrimitiveTypes = Union[str, int, float, bool]
 PrimitiveTypes = str
 
--- a/cloud_pipelines_backend/database_ops.py
+++ b/cloud_pipelines_backend/database_ops.py
@@ -75,5 +75,11 @@ def migrate_db(db_engine: sqlalchemy.Engine):
     # Or we need to avoid calling the Index constructor.
 
     for index in bts.ExecutionNode.__table__.indexes:
-        if index.name == "ix_execution_node_container_execution_cache_key":
+        if index.name == bts.ExecutionNode._IX_EXECUTION_NODE_CACHE_KEY:
             index.create(db_engine, checkfirst=True)
+            break
+
+    for index in bts.PipelineRunAnnotation.__table__.indexes:
+        if index.name == bts.PipelineRunAnnotation._IX_ANNOTATION_RUN_ID_KEY_VALUE:
+            index.create(db_engine, checkfirst=True)
+            break
--- a/cloud_pipelines_backend/filter_query_sql.py
+++ b/cloud_pipelines_backend/filter_query_sql.py
@@ -1,0 +1,246 @@
+import base64
+import json
+from typing import Any, Final
+
+import sqlalchemy as sql
+
+from . import backend_types_sql as bts
+from . import errors
+from . import filter_query_models
+
+# ---------------------------------------------------------------------------
+# Page-token helpers
+# ---------------------------------------------------------------------------
+
+_PAGE_TOKEN_OFFSET_KEY: Final[str] = "offset"
+_PAGE_TOKEN_FILTER_KEY: Final[str] = "filter"
+_PAGE_TOKEN_FILTER_QUERY_KEY: Final[str] = "filter_query"
+
+
+def _encode_page_token(*, page_token_dict: dict[str, Any]) -> str:
+    return base64.b64encode(json.dumps(page_token_dict).encode("utf-8")).decode("utf-8")
+
+
+def _decode_page_token(*, page_token: str | None) -> dict[str, Any]:
+    return json.loads(base64.b64decode(page_token)) if page_token else {}
+
+
+def _resolve_filter_value(
+    *,
+    filter: str | None,
+    filter_query: str | None,
+    page_token: str | None,
+) -> tuple[str | None, str | None, int]:
+    """Decode page_token and return the effective (filter_value, filter_query_value, offset).
+
+    If a page_token is present, its stored values take precedence over the
+    raw parameters (the token carries resolved values forward across pages).
+    """
+    page_token_dict = _decode_page_token(page_token=page_token)
+    offset = page_token_dict.get(_PAGE_TOKEN_OFFSET_KEY, 0)
+    if page_token:
+        filter = page_token_dict.get(_PAGE_TOKEN_FILTER_KEY)
+        filter_query = page_token_dict.get(_PAGE_TOKEN_FILTER_QUERY_KEY)
+    return filter, filter_query, offset
+
+
+# ---------------------------------------------------------------------------
+# Public API
+# ---------------------------------------------------------------------------
+
+
+def build_list_filters(
+    *,
+    filter_value: str | None,
+    filter_query_value: str | None,
+    page_token_value: str | None,
+    current_user: str | None,
+    page_size: int,
+) -> tuple[list[sql.ColumnElement], int, str]:
+    """Resolve pagination token, legacy filter, and filter_query into WHERE clauses.
+
+    Returns (where_clauses, offset, next_page_token_encoded).
+    """
+    if filter_value and filter_query_value:
+        raise errors.ApiValidationError(
+            "Cannot use both 'filter' and 'filter_query'. Use one or the other."
+        )
+
+    filter_value, filter_query_value, offset = _resolve_filter_value(
+        filter=filter_value,
+        filter_query=filter_query_value,
+        page_token=page_token_value,
+    )
+
+    where_clauses, resolved_filter = _build_filter_where_clauses(
+        filter_value=filter_value,
+        current_user=current_user,
+    )
+
+    if filter_query_value:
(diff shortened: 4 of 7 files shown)


cloud_pipelines_backend/database_ops.py
```diff
@@ -77,3 +77,18 @@ def migrate_db(db_engine: sqlalchemy.Engine):
     for index in bts.ExecutionNode.__table__.indexes:
         if index.name == "ix_execution_node_container_execution_cache_key":
             index.create(db_engine, checkfirst=True)
+
+    # Covering index for annotation-based filter_query predicates.
+    # Column order: (key, pipeline_run_id, value) enables B-tree seek on key,
+    # ordered pipeline_run_id scan, and avoids table lookups for all predicates.
+    annotation_index = sqlalchemy.Index(
```
yuechao-qin: - Define the index in the backend_type_sql.py
- then use the above "check for table indexes" to create the index if found.

cloud_pipelines_backend/database_ops.py
```diff
@@ -77,3 +77,18 @@ def migrate_db(db_engine: sqlalchemy.Engine):
...
+        "ix_pipeline_run_annotation_key_run_id_value",
+        bts.PipelineRunAnnotation.key,
+        bts.PipelineRunAnnotation.pipeline_run_id,
+        bts.PipelineRunAnnotation.value,
```
yuechao-qin: make order (key, value, id)

as annotations space grows, for example there are all created_by, then we can search against values performant.

cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -0,0 +1,238 @@
...
+        )
+
+    page_token = PageToken.decode(page_token_value)
+    offset = page_token.offset
+    filter_value = page_token.filter if page_token_value else filter_value
+    filter_query_value = (
+        page_token.filter_query if page_token_value else filter_query_value
+    )
+
+    where_clauses, resolved_filter = _build_filter_where_clauses(
```
Volv-G, line 75: we should probably put this under filter_value check as well, there's no point in executing this method without it anyway

Volv-G: also we should probably create empty where_clauses before this call and append to it here. And produce empty resolved_filter if filter_value isn't provided

cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -0,0 +1,238 @@
...
+def _predicate_to_clause(
+    *,
+    predicate,
+) -> sql.ColumnElement:
+    match predicate:
+        case AndPredicate():
+            return sql.and_(
+                *[_predicate_to_clause(predicate=p) for p in predicate.and_]
+            )
+        case OrPredicate():
```
Volv-G: should we consider making this part of the model classes to have less mapping?
i.e. define methods "to_clause(*, predicate)" in each model class that would return corresponding sql clauses

cloud_pipelines_backend/filter_query_sql.py
yuechao-qin: I'm actually going to merge `filter` and `filter_value` in an upstream PR.  So if legacy `filter` is defined, it'll be transformed into a `filter_query` format to use to same logic.  I believe that'll handle all the suggestions your requesting here.

cloud_pipelines_backend/filter_query_sql.py
yuechao-qin: That's pretty genius.  I wanted to do that, but there were design concerns that mixing SQL with Pydantic would potentially cause some side effects.  I decided to keep the responsibilities separate.

cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -0,0 +1,232 @@
...
+from . import filter_query_models
+
+# ---------------------------------------------------------------------------
+# PageToken
+# ---------------------------------------------------------------------------
+
+
+@dataclasses.dataclass(kw_only=True)
+class PageToken:
+    offset: int = 0
```
Ark-kun: Will you be moving to cursor way of filtering? If you plan to make changes soon, this class should not be public. Otherwise it's a breaking change.
Maybe you should just make the whole module private. Like `_api_server_search_utils.py`. ("utils" is an anti-pattern, but I do not have a better idea yet.)

cloud_pipelines_backend/component_structures.py
```diff
@@ -54,7 +54,6 @@
 import pydantic.alias_generators
 from pydantic.dataclasses import dataclass as pydantic_dataclasses
 
-
```
Ark-kun, line 57: Black usually puts exactly 2 blank lines between all module-level attributes that are not consecutive.
Here it separate imports from the constants, type variables and classes.

cloud_pipelines_backend/backend_types_sql.py
```diff
@@ -64,7 +71,7 @@ def generate_unique_id() -> str:
 
 # # Needed to put a union type into DB
 # class SqlIOTypeStruct(_BaseModel):
-#     type: structures.TypeSpecType
```
Ark-kun, line 67: This indentation was correct (it's a class variable inside a class).

cloud_pipelines_backend/backend_types_sql.py
```diff
@@ -2,14 +2,21 @@
...
 import sqlalchemy as sql
 from sqlalchemy import orm
 from sqlalchemy.ext import mutable
 
 IdType: typing.TypeAlias = str
 
+IX_EXECUTION_NODE_CACHE_KEY: Final[str] = (
+    "ix_execution_node_container_execution_cache_key"
+)
+IX_ANNOTATION_RUN_ID_KEY_VALUE: Final[str] = (
```
Ark-kun: `_IX_PIPELINE_RUN_ANNOTATION_RUN_ID_KEY_VALUE`

Let's also move it close to the class.

cloud_pipelines_backend/backend_types_sql.py
```diff
@@ -2,14 +2,21 @@
...
-from typing import Any
+from typing import Any, Final
 
 import sqlalchemy as sql
 from sqlalchemy import orm
 from sqlalchemy.ext import mutable
 
 IdType: typing.TypeAlias = str
 
+IX_EXECUTION_NODE_CACHE_KEY: Final[str] = (
```
Ark-kun: Let's make it private: `_IX_EXECUTION_NODE_CACHE_KEY`
Let's move it close to the class.

cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -0,0 +1,232 @@
...
+# Predicate dispatch
+# ---------------------------------------------------------------------------
+
+
+def _predicate_to_clause(
+    *,
+    predicate,
+) -> sql.ColumnElement:
+    match predicate:
+        case filter_query_models.AndPredicate():
```
Ark-kun, line 169: I don't like that we construct all these different objects just for comparison.

You can do:
```py
    match type(predicate):
        case filter_query_models.AndPredicate:
```

tests/test_filter_query_sql.py
```diff
@@ -0,0 +1,324 @@
...
+        assert "key" in compiled
+        assert "EXISTS" in compiled.upper()
+
+    def test_value_equals(self):
+        fq = filter_query_models.FilterQuery.model_validate(
+            {"and": [{"value_equals": {"key": "team", "value": "ml-ops"}}]},
+        )
+        clause = filter_query_sql.filter_query_to_where_clause(filter_query=fq)
+        compiled = _compile(clause)
+        assert "ml-ops" in compiled
```
Ark-kun: Maybe we can assert the whole query in some of these tests? It would be nice to see the full queries here.
I guess, there is a possibility that this query could be brittle and change with new version of SqlAlchemy. But we can wait for that moment and then make tests more robust if that happens.

Ark-kun (changes requested):

cloud_pipelines_backend/backend_types_sql.py
```diff
@@ -2,14 +2,21 @@
...
+    "ix_execution_node_container_execution_cache_key"
+)
+IX_ANNOTATION_RUN_ID_KEY_VALUE: Final[str] = (
+    "ix_pipeline_run_annotation_run_id_key_value"
```
Ark-kun: `ix_pipeline_run_annotation_run_id_key_value`
-->
`ix_pipeline_run_annotation_pipeline_run_id_key_value`

Ark-kun: Please change before testing in staging.

cloud_pipelines_backend/backend_types_sql.py
yuechao-qin: This was changed when running the black formatter.  I changed it back to have indentation.

cloud_pipelines_backend/filter_query_sql.py
yuechao-qin: I removed this dataclass and using the original encode/decode/resolve functions.

I'm open to adding an "_" to this module if you think that's helpful.  I rather not do that change in this PR, but a top PR in the stack to make restacking easier.

cloud_pipelines_backend/filter_query_sql.py
yuechao-qin: There is no new construction actually here, reference https://peps.python.org/pep-0636/#adding-a-ui-matching-objects



Simpler example:

tests/test_filter_query_sql.py
yuechao-qin: Sorry for missing this.  I agree the whole query is more helpful.

I did that in the upstream PRs, but missed it here.

Ark-kun (approved):
Approved with comments.

cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -0,0 +1,246 @@
...
+        key, sep, value = part.partition(":")
+        if sep:
+            parsed_filter[key] = value
+        else:
+            parsed_filter.setdefault("_text", "")
+            parsed_filter["_text"] += part
+    return parsed_filter
+
+
+def _build_filter_where_clauses(
```
morgan-wowk, line 124: It would be great if the functions that are specifically related to pipeline runs, had naming that mentioned pipeline runs. Alternatively, if they were constructed in a way where the caller manages the resource-specific aspects.

I can see us having filters on other endpoints in the future and just noticed this function, and the module it's in, does not clearly indicate it's for pipeline runs:

```
            if value:
                where_clauses.append(bts.PipelineRun.created_by == value)
            else:
                where_clauses.append(bts.PipelineRun.created_by == None)
```

cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Generic annotation-based leaf predicates
+# ---------------------------------------------------------------------------
+
+
+def _annotation_exists(*, conditions: list[sql.ColumnElement]) -> sql.ColumnElement:
+    return sql.exists(
+        sql.select(bts.PipelineRunAnnotation.pipeline_run_id).where(
```
morgan-wowk, line 199: Same feedback as above. We have pipeline run specific code in a module and function that doesn't clearly indicate it is for pipeline runs.

morgan-wowk: This pattern continues below.

cloud_pipelines_backend/filter_query_sql.py
yuechao-qin: I plan on doing this in the future when I do `Components Search`.  I didn't want to be too specific with any search feature naming right now, since I'm not sure what will overlap and what will be unique yet.  I'm hoping there are enough generic tools/features with `Pipeline Search` that can be used in other search too.

Is it ok if we defer this to the future when there are more use cases to better come up with proper naming mechanics?

cloud_pipelines_backend/filter_query_sql.py
yuechao-qin: Same as previous repsonse.

cloud_pipelines_backend/filter_query_sql.py
morgan-wowk: For sure. To me it was also an either now or when we start building out our next search use case. The latter is all good with me. Glad it's on your mind as well!

PR #106, opened 2026-02-24
feat: Create Pydantic models for Search Pipeline Run API

yuechao-qin:
### TL;DR

Added a new `filter_query` parameter to the pipeline runs API with Pydantic validation models, while maintaining backward compatibility with the existing `filter` parameter.

### What's new?

JSON definition for `filter_query`:

```json
{
  // logical (any predicate)
  "<and|or>": [
    // leaf
    {"key_exists": {"key": "<KEY>"}},
    // leaf
    {"value_contains": {"key": "<KEY>", "value_substring": "<VALUE>"}},
    // leaf
    {"value_in": {"key": "<KEY>", "values": ["<V1>", "<V2>", "..."]}},
    // leaf
    {"value_equals": {"key": "<KEY>", "value": "<VALUE>"}},
    // leaf
    {"time_range": {"key": "system/pipeline_run.date.created_at", "start_time": "<START_DATE>", "end_time": "<END_DATE>"}},
    // logical (leaf only)
    {"not": {"<LEAF_PREDICATE>": {}}},
    // logical (any predicate)
    {"<and|or>": [...]},
    ...
  ]
}
``` 

Example JSON for `filter_query`:

```json
{
  "and": [
    {"key_exists": {"key": "team"}},
    {"value_equals": {"key": "env", "value": "prod"}},
    {"value_in": {"key": "region", "values": ["us-east", "us-west", "eu-west"]}},
    {"not": {"key_exists": {"key": "deprecated"}}},
    {"or": [{"value_contains": {"key": "name", "value_substring": "nightly"}}]}
  ]
}
```

### What changed?

API `GET /api/pipeline_runs/`

#### Functional

- New `filter_query` if used in API will return an HTTP 501 (unimplemented error)
- Only `filter` or `filter_query` (mutual exclusive) can be in the API, else it returns a HTTP 422.

#### Other

- Added a new `filter_query` parameter to `ListPipelineRunsParams` that accepts structured JSON queries
- Created comprehensive Pydantic models in `filter_query_models.py` for validating filter queries, including:
  - Leaf predicates: `key_exists`, `value_equals`, `value_contains`, `value_in`, `time_range`
  - Logical operators: `and`, `or`, `not`
  - Recursive nesting support for complex queries
- Implemented JSON validation for `filter_query` with proper error handling
- Enforced strict validation rules including non-empty strings and timezone-aware datetimes

### How to test?

```bash
uv run pytest tests/test_api_server_sql.py tests/test_filter_query_models.py
```

[...]

PR #105, opened 2026-02-24
refactor: Search pipeline run refactoring and baseline unit tests

yuechao-qin:
### TL;DR

Refactored the `list` method in `PipelineRunsApiService_Sql` to improve code organization and added comprehensive test coverage for pipeline run operations.

#### Functional

None.

#### Other

- Extracted filter parsing logic into separate helper functions `_resolve_filter_value` and `_build_filter_where_clauses`
- Moved pipeline run response creation logic into a dedicated `_create_pipeline_run_response` method
- Added constants `_PAGE_TOKEN_OFFSET_KEY` and `_PAGE_TOKEN_FILTER_KEY` for page token field names
- Added comprehensive test suite covering listing, filtering, pagination, annotation CRUD operations, and helper functions

### How to test?

```
uv run pytest tests/test_api_server_sql.py
```

Run the new test suite in `tests/test_api_server_sql.py` which includes:
- Tests for listing pipeline runs with various filters and pagination
- Tests for pipeline name resolution and execution stats inclusion
- Tests for annotation management (set, delete, list)
- Tests for filter parsing and page token handling
- Edge cases like empty results and unsupported filters

### Why make this change?

1. **Refactoring upfront** — structural changes land here so upstream PR diffs only show new functionality, not mixed refactoring noise.
1. **Baseline tests** — establishing coverage now means each upstream PR's test additions map directly to its new features, making review clearer.

PR #108, opened 2026-02-24
feat: Search created by user in pipeline run API

yuechao-qin:
### TL;DR

Implemented search `created_by` user in Pipeline Runs.

#### Functionality

  - Search `created_by` user in `filter_query`.  Example query (not URL encoded for example):
    ```json
    /api?filter_query={"and": [{"value_equals": {"key": "system/pipeline_run.created_by", "value": "alice@example.com"}}]}
    ```
  - Annotation value `me` is supported for `created_by` search.

- API `POST /api/pipeline_runs/`
  - Copies the `created_by` data to the annotations table for future searching. Example pipeline run annotations table after API is called:
    | pipeline_run_id | key | value |
    |---|---|---|
    | `abc123` | `system/pipeline_run.created_by` | `alice@example.com` |

- API `[PUT|DELETE] /api/pipeline_runs/{id}/annotations/{key}`
  - Prevent (create, update, delete) of annotations with `system/` prefix.  Reserved for system annotation usage.

#### Other

- **Database migration**: Added backfill logic to populate existing pipeline runs with `created_by` annotations.  Example what a pipeline run's created by user would look like in the annotations table:

| pipeline_run_id | key | value |
|-----------------|-----|-------|
| 42 | system/pipeline_run.created_by | alice@example.com |

### How to test?

```
uv run pytest tests/test_api_server_sql.py  tests/test_filter_query_sql.py tests/test_database_ops.py
```

1. Create pipeline runs with different `created_by` values
2. Use filter queries like `{"and": [{"value_equals": {"key": "system/pipeline_run.created_by", "value": "alice"}}]}` to search by creator
3. Test the `"me"` placeholder: `{"and": [{"value_equals": {"key": "system/pipeline_run.created_by", "value": "me"}}]}`
4. Verify that attempts to set/delete system annotations return 422 errors
5. Test that unsupported predicates on system keys (like `value_contains`) are rejected

### Why make this change?

- This enables users to search for pipeline runs by creator using the new filter query system.
- Following safety guards and synchronization will allow old and new data (`create_by`) to be searchable with the new filter.
  - Preventing (create, delete, update) system prefix in annotations
  - Saving `created_by` to annotations table when starting a Pipeline Run

### DIFF
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -68,6 +68,14 @@ class ListPipelineJobsResponse:
 class PipelineRunsApiService_Sql:
     _PIPELINE_NAME_EXTRA_DATA_KEY = "pipeline_name"
     _DEFAULT_PAGE_SIZE: Final[int] = 10
+    _SYSTEM_KEY_RESERVED_MSG = (
+        "Annotation keys starting with "
+        f"{filter_query_sql.SYSTEM_KEY_PREFIX!r} are reserved for system use."
+    )
+
+    def _fail_if_changing_system_annotation(self, *, key: str) -> None:
+        if key.startswith(filter_query_sql.SYSTEM_KEY_PREFIX):
+            raise errors.ApiValidationError(self._SYSTEM_KEY_RESERVED_MSG)
 
     def create(
         self,
@@ -105,6 +113,19 @@ class PipelineRunsApiService_Sql:
                 },
             )
             session.add(pipeline_run)
+            # Mirror created_by into the annotations table so it's searchable
+            # via filter_query like any other annotation.
+            if created_by is not None:
+                # Flush to populate pipeline_run.id (server-generated) before inserting the annotation FK.
+                # TODO: Use ORM relationship instead of explicit flush + manual FK assignment.
+                session.flush()
+                session.add(
+                    bts.PipelineRunAnnotation(
+                        pipeline_run_id=pipeline_run.id,
+                        key=filter_query_sql.PipelineRunAnnotationSystemKey.CREATED_BY,
+                        value=created_by,
+                    )
+                )
             session.commit()
 
         session.refresh(pipeline_run)
@@ -295,6 +316,7 @@ class PipelineRunsApiService_Sql:
         user_name: str | None = None,
         skip_user_check: bool = False,
     ):
+        self._fail_if_changing_system_annotation(key=key)
         pipeline_run = session.get(bts.PipelineRun, id)
         if not pipeline_run:
             raise errors.ItemNotFoundError(f"Pipeline run {id} not found.")
@@ -317,6 +339,7 @@ class PipelineRunsApiService_Sql:
         user_name: str | None = None,
         skip_user_check: bool = False,
     ):
+        self._fail_if_changing_system_annotation(key=key)
         pipeline_run = session.get(bts.PipelineRun, id)
         if not pipeline_run:
             raise errors.ItemNotFoundError(f"Pipeline run {id} not found.")
--- a/cloud_pipelines_backend/database_ops.py
+++ b/cloud_pipelines_backend/database_ops.py
@@ -1,6 +1,8 @@
 import sqlalchemy
+from sqlalchemy import orm
 
 from . import backend_types_sql as bts
+from . import filter_query_sql
 
 
 def create_db_engine_and_migrate_db(
@@ -83,3 +85,60 @@ def migrate_db(db_engine: sqlalchemy.Engine):
         if index.name == bts.PipelineRunAnnotation._IX_ANNOTATION_RUN_ID_KEY_VALUE:
             index.create(db_engine, checkfirst=True)
             break
+
+    _backfill_pipeline_run_created_by_annotations(db_engine=db_engine)
+
+
+def _is_pipeline_run_annotation_key_already_backfilled(
+    *,
+    session: orm.Session,
+    key: str,
+) -> bool:
+    """Return True if at least one annotation with the given key exists."""
+    return session.query(
+        sqlalchemy.exists(
+            sqlalchemy.select(sqlalchemy.literal(1))
+            .select_from(bts.PipelineRunAnnotation)
+            .where(
+                bts.PipelineRunAnnotation.key == key,
+            )
+        )
+    ).scalar()
+
+
+def _backfill_pipeline_run_created_by_annotations(
+    *,
+    db_engine: sqlalchemy.Engine,
+) -> None:
+    """Copy pipeline_run.created_by into pipeline_run_annotation so
+    annotation-based search works for created_by.
+
+    The check and insert run in a single session/transaction to avoid
+    TOCTOU races between concurrent startup processes.
+
+    Skips entirely if any created_by annotation key already exists (i.e. the
+    write-path is populating them, so the backfill has already run or is
+    no longer needed).
+    """
+    with orm.Session(db_engine) as session:
+        if _is_pipeline_run_annotation_key_already_backfilled(
+            session=session,
+            key=filter_query_sql.PipelineRunAnnotationSystemKey.CREATED_BY,
+        ):
+            return
+
+        stmt = sqlalchemy.insert(bts.PipelineRunAnnotation).from_select(
+            ["pipeline_run_id", "key", "value"],
+            sqlalchemy.select(
+                bts.PipelineRun.id,
+                sqlalchemy.literal(
+                    filter_query_sql.PipelineRunAnnotationSystemKey.CREATED_BY
+                ),
+                bts.PipelineRun.created_by,
+            ).where(
+                bts.PipelineRun.created_by.isnot(None),
+                bts.PipelineRun.created_by != "",
+            ),
+        )
+        session.execute(stmt)
+        session.commit()
--- a/cloud_pipelines_backend/filter_query_models.py
+++ b/cloud_pipelines_backend/filter_query_models.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import abc
 from typing import Annotated
 
 import pydantic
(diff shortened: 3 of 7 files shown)


yuechao-qin:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/TangleML/tangle/108?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#108** <a href="https://app.graphite.com/github/pr/TangleML/tangle/108?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/108?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#107** <a href="https://app.graphite.com/github/pr/TangleML/tangle/107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#106** <a href="https://app.graphite.com/github/pr/TangleML/tangle/106?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#105** <a href="https://app.graphite.com/github/pr/TangleML/tangle/105?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

cloud_pipelines_backend/database_ops.py
```diff
@@ -83,3 +85,34 @@ def migrate_db(db_engine: sqlalchemy.Engine):
         if index.name == bts.IX_ANNOTATION_RUN_ID_KEY_VALUE:
             index.create(db_engine, checkfirst=True)
             break
+
+    backfill_created_by_annotations(db_engine)
+
+
+def backfill_created_by_annotations(db_engine: sqlalchemy.Engine):
```
yuechao-qin: For all backfills, check if key exists in any row to continue the backfill, else exit immediately.

cloud_pipelines_backend/database_ops.py
```diff
@@ -83,3 +85,52 @@ def migrate_db(db_engine: sqlalchemy.Engine):
         if index.name == bts.IX_ANNOTATION_RUN_ID_KEY_VALUE:
             index.create(db_engine, checkfirst=True)
             break
+
+    backfill_created_by_annotations(db_engine=db_engine)
+
+
+def is_annotation_key_already_backfilled(
```
Ark-kun: 1. It's a public function. Let's make it private.
2. The name mentions "annotation", but that work is used everywhere. We need to be more specific.

`_is_pipeline_run_annotation_key_already_backfilled`

cloud_pipelines_backend/database_ops.py
```diff
@@ -83,3 +85,52 @@ def migrate_db(db_engine: sqlalchemy.Engine):
...
+                sqlalchemy.select(sqlalchemy.literal(1))
+                .select_from(bts.PipelineRunAnnotation)
+                .where(
+                    bts.PipelineRunAnnotation.key == key,
+                )
+            )
+        ).scalar()
+
+
+def backfill_created_by_annotations(*, db_engine: sqlalchemy.Engine):
```
Ark-kun: 1. It's a public function. Let's make it private.
2. The name mentions "annotation", but that work is used everywhere. We need to be more specific.

`_backfill_pipeline_run_created_by_annotations`

cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -1,13 +1,30 @@
...
 import sqlalchemy as sql
 
 from . import backend_types_sql as bts
 from . import errors
 from . import filter_query_models
 
+SYSTEM_KEY_PREFIX: Final[str] = "system/"
+
+
+class SystemKey(enum.StrEnum):
```
Ark-kun: `PipelineRunAnnotationSystemKey`

cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -31,6 +48,93 @@ def decode(cls, token: str | None) -> "PageToken":
...
 
+# ---------------------------------------------------------------------------
+# SystemKey validation and resolution
+# ---------------------------------------------------------------------------
+
+
+def _get_predicate_key(*, predicate: filter_query_models.Predicate) -> str | None:
+    """Extract the annotation key from a leaf predicate, or None for logical operators."""
+    match predicate:
+        case filter_query_models.KeyExistsPredicate():
```
Ark-kun: You can have some base class like `filter_query_models.KeyPredicateBase` with field `key`. Then getting the key from the predicates would be trivial.

cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -149,17 +261,35 @@ def _build_filter_where_clauses(
 
 def _predicate_to_clause(
     *,
-    predicate,
+    predicate: filter_query_models.Predicate,
+    current_user: str | None = None,
```
Ark-kun, line 263: Maybe it's simpler to first preprocess the predicates so that you don't need to do that in functions like `_predicate_to_clause` ?

tests/test_database_ops.py
```diff
@@ -0,0 +1,271 @@
...
+                    container=structures.ContainerSpec(image="test-image")
+                ),
+            )
+        )
+    )
+
+
+@pytest.fixture()
+def session_factory():
+    engine = sqlalchemy.create_engine(
```
Ark-kun: Let's use the create_db_engine function from database_ops (without migration).

tests/test_database_ops.py
```diff
@@ -0,0 +1,271 @@
...
+
+def _create_run(session_factory, service, **kwargs):
+    """Create a pipeline run using a fresh session (mirrors production per-request sessions)."""
+    with session_factory() as session:
+        return service.create(session, **kwargs)
+
+
+class TestIsAnnotationKeyAlreadyBackfilled:
+    def test_false_on_empty_db(self, session_factory):
+        engine = session_factory.kw["bind"]
```
Ark-kun: Maybe, is_annotation_key_already_backfilled can just accept session_factory instead of engine?

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -297,6 +311,8 @@ def set_annotation(
         user_name: str | None = None,
         skip_user_check: bool = False,
     ):
+        if key.startswith(filter_query_sql.SYSTEM_KEY_PREFIX):
+            raise errors.InvalidAnnotationKeyError(_SYSTEM_KEY_RESERVED_MSG)
```
Ark-kun: Maybe, `self._fail_if_changing_system_annotation(key)`?

tests/test_database_ops.py
```diff
@@ -0,0 +1,271 @@
+import pytest
```
yuechao-qin, line 3: Add tests that indexes in migrate_db are being created.

cloud_pipelines_backend/filter_query_sql.py
yuechao-qin: Sorry, I'm not following?  If it's simple to give me an example of what you mean, that would be appreciated. Else, we can discuss this in person in our next 1-1.

tests/test_database_ops.py
yuechao-qin: That's a better design to make sure the check and backfill are within a single transaction session. Thanks.

cloud_pipelines_backend/filter_query_sql.py
Ark-kun: The idea was: have something like "_resolve_current_user_in_predicates". The functions like `_predicate_to_clause` won't need to bother with that. But if that's more complex, it's OK to leave things as they are.

cloud_pipelines_backend/filter_query_models.py
```diff
@@ -58,21 +59,45 @@ def _at_least_one_time_bound(self) -> TimeRange:
 # --- Predicate wrapper models (one field each) ---
 
 
-class KeyExistsPredicate(_BaseModel):
+class KeyPredicateBase(_BaseModel):
+    """Base for predicates that target an annotation key."""
+
+    @property
+    @abstractmethod
+    def key(self) -> str: ...
```
Ark-kun, line 67: Let's try to just use

`key: str`?

Then the derived classes should have it automatically (no need to override with a property.)

AFAIK, key cannot be a property, since that might not be recognized by Pydantic.

cloud_pipelines_backend/filter_query_models.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from abc import abstractmethod
```
morgan-wowk: question: I assume these standard library imports are an exception to the preference to import modules? CC: @Ark-kun

cloud_pipelines_backend/filter_query_models.py
morgan-wowk: The `key: str` solution might not work. If I'm understanding correctly, that would add `key` as a sibling to `value_contains` in the JSON, when it should be part of the nested object instead.

```json
{
  "value_contains": {
    "key": "...", # intended
    "value_substring": "..."
  },
  "key": "..." # not intended
}
```

This might help. Claude is suggesting Generics could be used, at the cost of having to understand Generics to read the code.

```python
from typing import Generic, Protocol, TypeVar
class HasKey(Protocol):
    key: str
T = TypeVar("T", bound=HasKey)
class KeyPredicateBase(_BaseModel, Generic[T]):
    # Internal name is always `payload`, JSON name is controlled per-subclass
    payload: T
    @property
    def key(self) -> str:
        return self.payload.key  # type checker knows T has .key
class ValueContainsPredicate(KeyPredicateBase[ValueContains]):
    payload: ValueContains = pydantic.Field(alias="value_contains")
class ValueEqualsPredicate(KeyPredicateBase[ValueEquals]):
    payload: ValueEquals = pydantic.Field(alias="value_equals")
```

yuechao-qin: @Ark-kun I tested just using 

`key: str`

but it doesn't work because:
- Each child predicate has a different path to `key` field.
- Property is intentional here, not to be a field. It's a convenience accessor to read the key uniformly across all predicate types without knowing which inner model they're using.
- If we do `key` in the base class as a field, Pydantic will expect `key` in the JSON filter.

cloud_pipelines_backend/filter_query_models.py
yuechao-qin: Great catch @morgan-wowk, thanks!

PR #120, opened 2026-02-25
Adds ruff and mypy linting and type checking

Mbeaulne:
## Summary

- Replace Black with Ruff for linting and formatting.

## Why Ruff over Black?

Ruff is a single tool that replaces Black (formatting), flake8 (linting), isort (import sorting), and pyupgrade (syntax modernization) -- all in one. It's written in Rust and runs 10-100x faster than the tools it replaces. Rather than maintaining separate configs and dependencies for each tool, Ruff handles everything through a single `[tool.ruff]` section in `pyproject.toml`.

To test this tool run the following command (note: all linting fixes are in the PR above):

`uv run ruff check --fix . && uv run ruff format .`


Mbeaulne:
* **#120** <a href="https://app.graphite.com/github/pr/TangleML/tangle/120?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/120?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

pyproject.toml
```diff
@@ -19,14 +19,36 @@ dependencies = [
 
 [dependency-groups]
 dev = [
-    "black>=26.1.0",
+    "mypy>=1.15",
```
Mbeaulne: This will be addressed more further up the stack

morgan-wowk:
Perhaps :point_up: will lead to a much shorter change to the `uv.lock` if we can eliminate whatever decided to do that change

morgan-wowk:
Would you mind putting together some instructions for how developers should set up their editor to use ruff instead of black?  
  
Right now, if I right click a document and select "Format Document" my understanding is that it formats the document using Black. I also have the cursor extension "Black Formatter" installed and assume after this myself and others would be recommended to uninstall it. That just leaves me wondering what other steps we would need to take to fully adopt ruff in our editor in place of black.

camielvs (approved):

PR #121, opened 2026-02-25
run ruff on codebase

Mbeaulne:
### TL;DR

Modernized Python code by adopting PEP 585 union syntax, consolidating imports, and applying various code style improvements.

### What changed?

- Replaced `typing.Union` and `typing.Optional` with modern union syntax (`|` and `| None`)
- Consolidated multiple import statements from the same module into single lines
- Removed unused imports (`Optional`, `OrderedDict`, etc.)
- Updated type annotations to use built-in generics (`list`, `dict`, `tuple`) instead of `typing` equivalents
- Replaced f-string formatting with direct string literals where variables weren't used
- Simplified conditional logic in some cases (combined elif conditions)
- Updated exception handling to remove unused exception variables
- Modernized file opening calls by removing redundant mode specifications
- Fixed import ordering to follow PEP 8 conventions

### How to test?

Run the existing test suite to ensure all functionality remains intact. The changes are purely stylistic and should not affect runtime behavior.

### Why make this change?

These changes improve code readability and maintainability while adopting modern Python conventions. The new union syntax is more concise and readable, consolidated imports reduce clutter, and removing unused imports helps with code clarity and potentially reduces import overhead.

### DIFF
--- a/cloud_pipelines_backend/launchers/kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/kubernetes_launchers.py
@@ -7,7 +7,7 @@ import logging
 import os
 import pathlib
 import typing
-from typing import Any, Optional
+from typing import Any
 
 from kubernetes import client as k8s_client_lib
 from kubernetes import watch as k8s_watch_lib
@@ -17,9 +17,9 @@ from cloud_pipelines.orchestration.storage_providers import (
     interfaces as storage_provider_interfaces,
 )
 from cloud_pipelines.orchestration.storage_providers import local_storage
+
 from .. import component_structures as structures
-from . import container_component_utils
-from . import interfaces
+from . import container_component_utils, interfaces
 
 if typing.TYPE_CHECKING:
     from google.cloud import storage
@@ -417,7 +417,7 @@ class _KubernetesPodLauncher(
         output_uris: dict[str, str],
         log_uri: str,
         annotations: dict[str, Any] | None = None,
-    ) -> "LaunchedKubernetesContainer":
+    ) -> LaunchedKubernetesContainer:
         namespace = self._choose_namespace(annotations=annotations)
 
         pod = self._prepare_kubernetes_pod(
@@ -466,7 +466,7 @@ class _KubernetesPodLauncher(
 
     def get_refreshed_launched_container_from_dict(
         self, launched_container_dict: dict
-    ) -> "LaunchedKubernetesContainer":
+    ) -> LaunchedKubernetesContainer:
         launched_container = LaunchedKubernetesContainer.from_dict(
             launched_container_dict, launcher=self
         )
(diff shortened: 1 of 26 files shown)


morgan-wowk:
Seems pretty good to me. Can you ask an agent if any of the changes made in the most recent commit (assuming you are checked out on this branch) are breaking changes? I'd be interested to have that peace of mind, especially given my own Python knowledge is limited and I won't be able to properly judge line by line if this is safe.  
  
If you think it's safe, and an agent thinks it's safe, and a regression test of the API and orchestrators passes locally, then it's good with me!

Ark-kun:
Thank you for your work on improving the codebase.

There couple of notes:

>run ruff

We use Black for code formatting.

Our style guide is based on Google Style guide: https://google.github.io/styleguide/pyguide.html with certain exception like using relative imports.

>adopting PEP 585 union syntax

Note that the type union syntax is https://peps.python.org/pep-0604/ which was only introduced in Python 3.10.
Since Python 3.9 is EOL, I guess we can drop support. But usually we try to support users unless there is some compelling feature that we want.

Let's make sure the resulting changes work on Python 3.10 minimum version. There are some delayed type annotations that I think are only supported in later Python versions (for example, referring to the function's enclosing class).

We also use https://www.conventionalcommits.org/ . This will make it easy to generate changelogs.

cloud_pipelines_backend/launchers/google_kubernetes_launchers.py
```diff
@@ -1,8 +1,7 @@
 import typing
 
-from kubernetes import client as k8s_client_lib
-
 from cloud_pipelines.orchestration.storage_providers import google_cloud_storage
```
Ark-kun: We usually put cloud_pipelines import in a group adjacent to cloud_pipelines_backend since these are closely related projects.

cloud_pipelines_backend/launchers/kubernetes_launchers.py
```diff
@@ -582,7 +581,7 @@ def __init__(
         service_account_name: str | None = None,
         request_timeout: int | tuple[int, int] = 10,
         pod_name_prefix: str = "task-pod-",
-        gcs_client: "storage.Client | None" = None,
+        gcs_client: storage.Client | None = None,
```
Ark-kun, line 585: In which versions of Python will this work?

camielvs (approved):

uv.lock
```diff
@@ -9,7 +9,7 @@ resolution-markers = [
 [[package]]
 name = "annotated-types"
 version = "0.7.0"
-source = { registry = "https://pypi.org/simple" }
+source = { registry = "https://pkgs.shopify.io/basic/data/python/simple/" }
```
Ark-kun, line 12: We should not use Shopify's private package repository for OSS.

PR #123, opened 2026-02-25
chore: Refactor OTel concerns in preparation for metric provider

morgan-wowk:
## Refactor OpenTelemetry instrumentation into modular package

**Changes:**

* Restructured OpenTelemetry code into a dedicated package with clear public API
* Split provider setup and FastAPI instrumentation into separate functions (`setup_providers()` and `instrument_fastapi()`)
* Added configuration validation and error handling for OpenTelemetry settings
* Implemented provider state checking to conditionally enable instrumentation only when providers are configured
* Created internal modules for shared configuration resolution and provider state management
* Added comprehensive test coverage for all new modules and functions
* Updated main application files to use the new simplified API calls


cloud_pipelines_backend/instrumentation/opentelemetry/_internal/configuration.py
```diff
@@ -0,0 +1,53 @@
...
+    Raises ValueError if the configuration is invalid.
+    """
+    otel_endpoint = os.environ.get("TANGLE_OTEL_EXPORTER_ENDPOINT")
+    if not otel_endpoint:
+        return None
+
+    otel_protocol = os.environ.get("TANGLE_OTEL_EXPORTER_PROTOCOL", _OTEL_PROTOCOL_GRPC)
+
+    if service_name is None:
+        app_env = os.environ.get("TANGLE_ENV", "unknown")
```
yuechao-qin, line 39: Trying to understand the effect of "unknown" as a service name default.

would this result in alot of unknowns service names in OpenTelemetry, which means a bucket of unknowns for different services.  Is that what we want or a UID to distinguish different unknown services?

tests/instrumentation/opentelemetry/_internal/conftest.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+from opentelemetry import trace
+
+
+@pytest.fixture(autouse=True)
+def reset_otel_tracer_provider():
```
yuechao-qin, line 6: nit: would it make sense to only have 1 of this and use it in all the tests?

cloud_pipelines_backend/instrumentation/opentelemetry/providers.py
```diff
@@ -0,0 +1,42 @@
...
+    if otel_config is None:
+        return
+
+    tracing.setup(
+        endpoint=otel_config.endpoint,
+        protocol=otel_config.protocol,
+        service_name=otel_config.service_name,
+    )
+
+    # TODO: Setup metrics provider once it's available
```
yuechao-qin, line 42: is this for an upstream PR?

cloud_pipelines_backend/instrumentation/opentelemetry/_internal/config.py
```diff
@@ -0,0 +1,53 @@
...
+Shared OpenTelemetry configuration resolution.
+
+Reads and validates OTel settings from environment variables.
+"""
+
+import dataclasses
+import os
+
+_OTEL_PROTOCOL_GRPC = "grpc"
+_OTEL_PROTOCOL_HTTP = "http"
```
yuechao-qin: Use a StrEnum so you can reference it like this in other files (below) and it groups all protocols together in a structure.

```
if protocol == OtelProtocols.GPRC:
```

You don't have to create the tuple either `_OTEL_PROTOCOLS`, you can just do a check like this `"SOME_STRING" in <MyStrEnumObject>`

cloud_pipelines_backend/instrumentation/opentelemetry/_internal/configuration.py
morgan-wowk: This just acts as a flag to us to set the environment variable. For example, if we see `unknown` in the observability platform then it's an indicator to us to set that environment variable.

morgan-wowk: This is the best practice in this case. The team in control of the deployment would be responsible for setting this. They might experience a short time where they are getting data emitted to `tangle-unknown` that will inevitably be dropped after the retention period.

tests/instrumentation/opentelemetry/_internal/conftest.py
morgan-wowk: Unfortunately this is not possible at the root level or recommended in general. I debated this with the agent and we came to the conclusion that it is worth the duplication in this case. It keeps packages independent of each other, and also prevents us from having test-related files outside the tests folder.

cloud_pipelines_backend/instrumentation/opentelemetry/providers.py
morgan-wowk: Yuppers! 2nd from the top

cloud_pipelines_backend/instrumentation/opentelemetry/_internal/config.py
morgan-wowk: I dig it

morgan-wowk: Thanks!

cloud_pipelines_backend/instrumentation/opentelemetry/_internal/config.py
```diff
@@ -0,0 +1,53 @@
...
+_OTEL_PROTOCOLS = (_OTEL_PROTOCOL_GRPC, _OTEL_PROTOCOL_HTTP)
+
+
+@dataclasses.dataclass(frozen=True)
```
yuechao-qin: I like verbosity in code, so I side with doing this with my dataclasses

```
@dataclass(kw_only=True)
class Config:
    name: str
    value: int
    enabled: bool = True

# Must use keyword arguments
Config(name="foo", value=42) # OK
Config("foo", 42) # TypeError
```

Highly suggest this, for this PR and upstream ones too.

morgan-wowk: Adopted through the entire stack. Thank you!

start_local.py
```diff
@@ -132,6 +132,12 @@ def get_user_details(request: fastapi.Request):
 logger = logging.getLogger(__name__)
 # endregion
 
+# region: OpenTelemetry initialization
+from cloud_pipelines_backend.instrumentation import opentelemetry as otel
+
+otel.setup_providers()
```
maxy-shpfy, line 138: Question: for local setup, should otel service be run separately? What would happen if no OTEL endpoint is configured (e.g. in case of e2e tests on CI)?

morgan-wowk: Great question.  
  
I'll answer in two ways.  
  
For local development, we have this example stack I added and have personally been using to test traces and metrics for OSS: https://github.com/TangleML/tangle/tree/master/examples/observability/otel-jaeger-prometheus

For writing tests, or adding new metrics, here are some important things to know:

- The OTel Python SDK that we consume, and even our FastAPI auto-instrumentation consumes, is intentionally designed so that every action is a NOOP if no providers are set.
    - It means, no exceptions from emitting metrics or opening traces when OTel is not configured.
- We only set providers on our end if they have filled out correctly shaped values for the environment variables.
- We only auto-instrument FastAPI if providers are set.
- For **verfiying** a metric, or trace, was used, via testing (whether its unit or e2e) that is possible and I'll follow up with instructions and examples in a separate PR. Consider this is a TODO. We will have to manipulate the type of reader / exporter we use at the start of tests.

morgan-wowk: Here is an issue: https://github.com/TangleML/tangle/issues/132

Ark-kun:
Let's keep the tests inside /tests/. We can have subfolders there like instrumentation or telemetry

cloud_pipelines_backend/instrumentation/opentelemetry/_internal/tests/test_config.py
```diff
@@ -0,0 +1,111 @@
...
+
+        assert result is not None
+        assert result.endpoint == "http://localhost:4317"
+        assert result.protocol == config.ExporterProtocol.GRPC
+        assert result.service_name == "tangle-unknown"
+
+    def test_uses_custom_service_name(self, monkeypatch):
+        monkeypatch.setenv("TANGLE_OTEL_EXPORTER_ENDPOINT", "http://localhost:4317")
+
+        result = config.resolve(service_name="oasis-api")
```
Ark-kun: We need to switch naming from Oasis to Tangle

cloud_pipelines_backend/instrumentation/opentelemetry/_internal/configuration.py
Ark-kun, line 1: Naming module with a singular noun often conflicts with local variables. Better use plural or more uncountable. (Maybe, `configuration` is OK despite technically being singular noun...)

Kubernetes has

```
from kubernetes import config
from kubernetes import client
```

and it's very inconvenient, forcing me to rename those imports every time.

morgan-wowk:
> Let's keep the tests inside /tests/. We can have subfolders there like instrumentation or telemetry

​Sure.

Any reason you want to avoid co-locating tests? It generally comes with many benefits:

- Increased discoverability
    - Promotes writing tests
    - Makes tests easy to find
- Faster to update tests
- Test proof
    - A clear indicator when a given package is tested, or missing tests
    - You don't have to search for tests in large folder of tests, or folder of folders
- Package isolation has some benefits itself.

Increasingly over time, our root `/tests` directory will become difficult to search through. I find that it is already becoming that way.

morgan-wowk:
Here is what I recommend:

End-to-end / integration tests in the root `/tests`

Unit tests co-located with the code they are testing.

cloud_pipelines_backend/instrumentation/opentelemetry/_internal/configuration.py
morgan-wowk: Fair! Changed.

cloud_pipelines_backend/instrumentation/opentelemetry/_internal/tests/test_config.py
morgan-wowk: Changed

Ark-kun:
Co-location tests has advantages, but also have several cons.
Tests get included in the installed package, problems with test data, imports etc.

Usually there is 1-1 mapping between modules and test modules:
```
my_modeule.py
my_modeule_tests.py
```
But when tests start to get their own `conftest.py` files, test data, utility modules, etc, I think it's definitely time for them to live in their own place.

pyproject.toml
```diff
@@ -27,6 +27,9 @@ huggingface = [
     "huggingface-hub[oauth]>=0.35.3",
 ]
 
+[tool.pytest.ini_options]
+testpaths = ["tests"]
+addopts = "--import-mode=importlib"
```
Ark-kun, line 32: Which errors required making this change?
Is this still needed?

morgan-wowk:
Good to know about this behaviour in Python. Thanks guys

pyproject.toml
morgan-wowk: I will run the tests again tomorrow and confirm. And paste the error if it is still present.

morgan-wowk: The answer is yes, it is still needed.  
  
**Context:**  
_`--import-mode=importlib`_ _is required because the test layout has two files with the same name:_ _`test_providers.py`_ _exists in both_ _`tests/instrumentation/opentelemetry/`_ _and_ _`tests/instrumentation/opentelemetry/_internal/`. Pytest's default_ _`prepend`_ _import mode can't distinguish between identically-named test modules in sibling directories — it errors with "which is not the same as the test file we want to collect"._ _`importlib`_ _mode gives each file a unique identity based on its full path rather than its module name, which resolves the collision._

PR #125, opened 2026-02-25
chore!: Refactor OTel exporters to be signal-specific

morgan-wowk:
## Refactor OpenTelemetry configuration to support signal-specific exporters

This refactors the OpenTelemetry configuration system to support separate exporter configurations for different telemetry signals (traces, metrics, logs).

### Changes

- Split `OtelConfig` into separate `ExporterConfig` and `OtelConfig` classes
- Moved exporter-specific fields (`endpoint`, `protocol`) to `ExporterConfig`
- Added `trace_exporter` field to `OtelConfig` for trace-specific configuration
- Updated environment variable names from generic `TANGLE_OTEL_EXPORTER_*` to signal-specific `TANGLE_OTEL_TRACE_EXPORTER_*`
- Extracted `_resolve_exporter()` helper function for reusable exporter configuration logic
- Modified `providers.setup()` to only configure tracing when trace exporter is present
- Updated all tests to use the new environment variable names and configuration structure

This allows users to configure different exporters for traces vs metrics by setting signal-specific environment variables.

### DIFF
--- a/cloud_pipelines_backend/instrumentation/opentelemetry/_internal/configuration.py
+++ b/cloud_pipelines_backend/instrumentation/opentelemetry/_internal/configuration.py
@@ -15,29 +15,60 @@ class ExporterProtocol(str, enum.Enum):
 
 
 @dataclasses.dataclass(frozen=True, kw_only=True)
-class OtelConfig:
+class ExporterConfig:
     endpoint: str
     protocol: str
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class OtelConfig:
     service_name: str
     service_version: str
+    trace_exporter: ExporterConfig | None = None
+
+
+def _resolve_exporter(
+    endpoint_var: str,
+    protocol_var: str,
+) -> ExporterConfig | None:
+    endpoint = os.environ.get(endpoint_var)
+    if not endpoint:
+        return None
+
+    protocol = os.environ.get(protocol_var, ExporterProtocol.GRPC)
+
+    if not endpoint.startswith(("http://", "https://")):
+        raise ValueError(
+            f"Invalid OTel endpoint format: {endpoint}. "
+            f"Expected format: http://<host>:<port> or https://<host>:<port>"
+        )
+    try:
+        ExporterProtocol(protocol)
+    except ValueError:
+        raise ValueError(
+            f"Invalid OTel protocol: {protocol}. "
+            f"Expected values: {', '.join(e.value for e in ExporterProtocol)}"
+        )
+
+    return ExporterConfig(endpoint=endpoint, protocol=protocol)
 
 
 def resolve(
     service_name: str | None = None,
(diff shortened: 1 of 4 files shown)


cloud_pipelines_backend/instrumentation/opentelemetry/_internal/configuration.py
```diff
@@ -15,53 +15,70 @@ class ExporterProtocol(str, enum.Enum):
...
 
+    if trace_exporter is None:
+        return None
+
     if service_name is None:
         app_env = os.environ.get("TANGLE_ENV", "unknown")
         service_name = f"tangle-{app_env}"
 
     if service_version is None:
         service_version = os.environ.get("TANGLE_SERVICE_VERSION", "unknown")
```
maxy-shpfy, line 78: Do we have this env set today? what the versioning strategy do we use?

cloud_pipelines_backend/instrumentation/opentelemetry/_internal/configuration.py
```diff
@@ -15,53 +15,70 @@ class ExporterProtocol(str, enum.Enum):
...
+    trace_exporter = _resolve_exporter(
+        endpoint_var="TANGLE_OTEL_TRACE_EXPORTER_ENDPOINT",
+        protocol_var="TANGLE_OTEL_TRACE_EXPORTER_PROTOCOL",
     )
 
+    if trace_exporter is None:
+        return None
+
     if service_name is None:
         app_env = os.environ.get("TANGLE_ENV", "unknown")
```
maxy-shpfy, line 74: What are possible values for TANGLE_ENV? Where it is being set?

morgan-wowk: We improve the documentation for this upstack but happy to answer here:  
  
This is used as part of the service name for your OTel exports. This is literally your env, so `development`, `staging`, etc.  
  
Alexey and I landed on this name for it in an earlier merged PR.

cloud_pipelines_backend/instrumentation/opentelemetry/_internal/configuration.py
morgan-wowk: Technically, it's up to the OSS user of Tangle (someone who clones and deploys Tangle themselves).  
  
They can use their commit shas, semantic versioning, whatever they feel is right. For our case, we use a revision (commit sha) provided by our CD platform.  
  
As for recommendations, we could add some to our documentation of this variable which is upstack. Suitable values would be build ids, URLs, or commit shas.

morgan-wowk: PS: I'm not fully happy with the docs yet even after this full stack as it is now. I'd love to collaborate and hear ideas on how in-depth we want to go on docs and where our source of truth should be for metric information.

PR #129, opened 2026-02-26
feat: Search pipeline name in pipeline run API

yuechao-qin:
### TL;DR

Implemented search `name` in Pipeline Runs.

#### Functional

  - Search pipeline run name in `filter_query`.  Example query (not URL encoded for example):
    ```json
    /api?filter_query={"and": [{"value_equals": {"key": "system/pipeline_run.name", "value": "My pipeline run name"}}]}
    ```

- API `POST /api/pipeline_runs/`
  - Copies the pipeline run name to the annotations table for future searching.  Example pipeline run annotations table after API is called:
    | pipeline_run_id | key | value |
    |---|---|---|
    | `abc123` | `system/pipeline_run.name` | `My pipeline run name` |

#### Other

- Implemented `backfill_pipeline_name_annotations` function to populate existing pipeline runs with name annotations

### How to test?

```
uv run pytest tests/test_api_server_sql.py tests/test_filter_query_sql.py tests/test_database_ops.py
```

- Create pipeline runs with various name configurations and verify that name annotations are properly created
- Test the backfill functionality by removing annotations and running the backfill process
- Verify that filter queries using the new `SystemKey.NAME` work with different predicate types (exists, equals, contains, in)
- Confirm that both user annotations and system annotations coexist properly

### Why make this change?

- This enables users to search and filter pipeline runs by name. 
- By mirroring pipeline names as system annotations, pipeline names will be searchable via the annotations table.

### DIFF
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -9,6 +9,7 @@ from sqlalchemy import orm
 
 from . import backend_types_sql as bts
 from . import component_structures as structures
+from . import database_ops
 from . import errors
 from . import filter_query_sql
 
@@ -30,6 +31,27 @@ class ApiServiceError(RuntimeError):
 
 def _get_current_time() -> datetime.datetime:
     return datetime.datetime.now(tz=datetime.timezone.utc)
+
+
+def _get_pipeline_name_from_task_spec(
+    *,
+    task_spec_dict: dict[str, Any],
+) -> str | None:
+    """Extract pipeline name from a task_spec dict via component_ref.spec.name.
+
+    Traversal path:
+        task_spec_dict -> TaskSpec -> component_ref -> spec -> name
+
+    Returns None if any step in the chain is missing or parsing fails.
+    """
+    try:
+        task_spec = structures.TaskSpec.from_json_dict(task_spec_dict)
+    except Exception:
+        return None
+    spec = task_spec.component_ref.spec
+    if spec is None:
+        return None
+    return spec.name or None
 
 
 # ==== PipelineJobService
@@ -113,19 +135,15 @@ class PipelineRunsApiService_Sql:
                 },
             )
             session.add(pipeline_run)
-            # Mirror created_by into the annotations table so it's searchable
-            # via filter_query like any other annotation.
-            if created_by is not None:
-                # Flush to populate pipeline_run.id (server-generated) before inserting the annotation FK.
-                # TODO: Use ORM relationship instead of explicit flush + manual FK assignment.
-                session.flush()
-                session.add(
-                    bts.PipelineRunAnnotation(
-                        pipeline_run_id=pipeline_run.id,
-                        key=filter_query_sql.PipelineRunAnnotationSystemKey.CREATED_BY,
-                        value=created_by,
-                    )
-                )
+            # Flush to populate pipeline_run.id (server-generated) before inserting annotation FKs.
+            # TODO: Use ORM relationship instead of explicit flush + manual FK assignment.
+            session.flush()
+            _mirror_system_annotations(
+                session=session,
+                pipeline_run_id=pipeline_run.id,
+                created_by=created_by,
+                pipeline_name=pipeline_name,
+            )
             session.commit()
 
         session.refresh(pipeline_run)
@@ -244,12 +262,9 @@ class PipelineRunsApiService_Sql:
                     bts.ExecutionNode, pipeline_run.root_execution_id
                 )
                 if execution_node:
-                    task_spec = structures.TaskSpec.from_json_dict(
-                        execution_node.task_spec
+                    pipeline_name = _get_pipeline_name_from_task_spec(
+                        task_spec_dict=execution_node.task_spec
                     )
-                    component_spec = task_spec.component_ref.spec
-                    if component_spec:
-                        pipeline_name = component_spec.name
             response.pipeline_name = pipeline_name
         if include_execution_stats:
             execution_status_stats = self._calculate_execution_status_stats(
@@ -1151,6 +1166,32 @@ class SecretsApiService:
 _ArtifactNodeOrDynamicDataType = typing.Union[
     bts.ArtifactNode, structures.DynamicDataArgument
 ]
+
+
+def _mirror_system_annotations(
+    *,
+    session: orm.Session,
+    pipeline_run_id: bts.IdType,
+    created_by: str | None,
+    pipeline_name: str | None,
+) -> None:
+    """Mirror pipeline run fields as system annotations for filter_query search."""
+    if created_by:
+        session.add(
+            bts.PipelineRunAnnotation(
+                pipeline_run_id=pipeline_run_id,
+                key=filter_query_sql.PipelineRunAnnotationSystemKey.CREATED_BY,
+                value=created_by,
+            )
+        )
+    if pipeline_name:
+        session.add(
+            bts.PipelineRunAnnotation(
+                pipeline_run_id=pipeline_run_id,
+                key=filter_query_sql.PipelineRunAnnotationSystemKey.PIPELINE_NAME,
+                value=pipeline_name,
+            )
+        )
 
 
 def _recursively_create_all_executions_and_artifacts_root(
--- a/cloud_pipelines_backend/database_ops.py
+++ b/cloud_pipelines_backend/database_ops.py
@@ -1,3 +1,5 @@
+from typing import Any
+
 import sqlalchemy
 from sqlalchemy import orm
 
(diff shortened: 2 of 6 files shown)


cloud_pipelines_backend/database_ops.py
```diff
@@ -87,6 +87,38 @@ def migrate_db(db_engine: sqlalchemy.Engine):
...
+
+
+def backfill_pipeline_name_annotations(db_engine: sqlalchemy.Engine):
+    """Copy pipeline_run.extra_data['pipeline_name'] into pipeline_run_annotation
+    so annotation-based search works for pipeline names.
+
+    Idempotent -- skips rows that already have the annotation.
+    """
+    with orm.Session(db_engine) as session:
+        pipeline_name_expr = bts.PipelineRun.extra_data["pipeline_name"].as_string()
```
yuechao-qin: Extra data and pipeline_name key might not exist, need to check both exists to safely traverse to get the name.

If nothing is found, then get the name by doing `pipeline_run.execution_node.task_spec.componentRef.spec.name` (taverse safely to get the name as there cal be null objects in the path).

tests/test_database_ops.py
```diff
@@ -103,112 +154,410 @@ def test_matches_exact_key(self, session_factory):
...
+            session_factory,
+            service,
+            root_task=_make_task_spec("some-name"),
         )
+        key = filter_query_sql.SystemKey.NAME
+        _delete_annotation(session_factory=session_factory, run_id=run.id, key=key)
 
+        with session_factory() as session:
+            db_run = session.get(bts.PipelineRun, run.id)
+            db_run.extra_data = {"pipeline_name": ""}
```
yuechao-qin, line 500: If string is empty "", trust that pipeline name is empty (no need to check in spec) and set pipeline name = "" in annotations table.

cloud_pipelines_backend/database_ops.py
```diff
@@ -134,3 +162,142 @@ def backfill_created_by_annotations(*, db_engine: sqlalchemy.Engine):
...
+        Python fallback for runs Phase 1 missed (extra_data is NULL or
+        missing the key). Resolves name via component_ref.spec.name.
+    """
+    if is_annotation_key_already_backfilled(
+        db_engine=db_engine, key=filter_query_sql.SystemKey.NAME
+    ):
+        return
+
+    _backfill_pipeline_names_from_extra_data(db_engine=db_engine)
+    _backfill_pipeline_names_from_component_spec(db_engine=db_engine)
```
yuechao-qin: Do these in a single transaction.

Also, look into removing phase 1 (extra_data) and see if we can do eveyrhting for phase 2 in a single query.

https://dev.mysql.com/doc/refman/8.4/en/json-search-functions.html

```
SELECT pipeline_run.id, "system/piepline_run.name", JSON_EXTRACT(execution_node.task_spec, ".componentRef.spec.name")
FROM pipeline_run
JOIN
execution_node

INSERT from SELECT
```

cloud_pipelines_backend/database_ops.py
```diff
@@ -142,3 +167,232 @@ def _backfill_pipeline_run_created_by_annotations(
...
+            existing_ann,
+            sqlalchemy.and_(
+                existing_ann.pipeline_run_id == bts.PipelineRun.id,
+                existing_ann.key == key,
+            ),
+        ).where(
+            # Step 2b: Anti-join — keep only runs with no existing annotation
+            existing_ann.pipeline_run_id.is_(None),
+            # Step 3: JSON extraction — keep only non-NULL names
+            name_expr.isnot(None),
```
Ark-kun, line 332: `.isnot(...)` still works after `.as_string()`?

cloud_pipelines_backend/database_ops.py
```diff
@@ -142,3 +167,232 @@ def _backfill_pipeline_run_created_by_annotations(
...
+
+    Any null at any depth (task_spec NULL, componentRef missing,
+    spec null, name missing) produces SQL NULL, filtered out by
+    IS NOT NULL. Empty string is allowed and will be inserted.
+    """
+    key = filter_query_sql.PipelineRunAnnotationSystemKey.NAME
+    name_expr = bts.ExecutionNode.task_spec[
+        ("componentRef", "spec", "name")
+    ].as_string()
+    existing_ann = orm.aliased(bts.PipelineRunAnnotation)
```
Ark-kun, line 306: Just curious: Why do we need aliasing? Is this beacause we're inserting in same table?

cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -10,10 +10,12 @@
 from . import filter_query_models
 
 SYSTEM_KEY_PREFIX: Final[str] = "system/"
+_PIPELINE_RUN_KEY_PREFIX: Final[str] = f"{SYSTEM_KEY_PREFIX}pipeline_run."
 
 
 class PipelineRunAnnotationSystemKey(enum.StrEnum):
-    CREATED_BY = f"{SYSTEM_KEY_PREFIX}pipeline_run.created_by"
+    CREATED_BY = f"{_PIPELINE_RUN_KEY_PREFIX}created_by"
+    NAME = f"{_PIPELINE_RUN_KEY_PREFIX}name"
```
Ark-kun: Maybe, `pipeline_name ?`

tests/test_database_ops.py
```diff
@@ -44,6 +99,23 @@ def _create_run(
...
+def _delete_annotation(
+    *,
+    session_factory: orm.sessionmaker,
+    run_id: str,
+    key: filter_query_sql.PipelineRunAnnotationSystemKey,
+) -> None:
+    """Remove a write-path annotation so backfill can be tested in isolation."""
+    with session_factory() as session:
+        session.execute(
+            sqlalchemy.delete(bts.PipelineRunAnnotation).where(
```
Ark-kun, line 111: It's OK in tests, but in other places let's always use `session.delete(session.get(Something, key1))` pattern instead of "delete where" queries ()when deleting a single row). This avoids deleting many rows by mistake.

cloud_pipelines_backend/database_ops.py
```diff
@@ -106,6 +110,27 @@ def _is_pipeline_run_annotation_key_already_backfilled(
     ).scalar()
 
 
+def get_pipeline_name_from_task_spec(
```
Ark-kun: It's only used in api_server_sql.py, so maybe let's move it there. (and make it private)

tests/test_database_ops.py
yuechao-qin: Thanks, I learned something new!

cloud_pipelines_backend/database_ops.py
yuechao-qin: Exactly, since we `INSERT` and do a `LEFT JOIN` on the same table in one query, the alias helps the DB to not be confused with the read reference and write target.

cloud_pipelines_backend/database_ops.py
yuechao-qin: Yes, `.as_string()` returns a SQLAlchemy `ColumnElement[str]` (not a Python string), which provides `.isnot()` function. It generates `IS NOT NULL` in the final SQL.

I also have tests which exercise every NULL scenario:
- `test_p2_depth1_task_spec_null` — task_spec is NULL
- `test_p2_depth1_task_spec_empty` — empty dict, key absent
- `test_p2_depth2_component_ref_missing` — componentRef missing
- `test_p2_depth2_component_ref_null` — componentRef is null
- `test_p2_depth3_spec_null` — spec is null
- `test_p2_depth4_name_null` — name is null

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -244,12 +241,9 @@ def _create_pipeline_run_response(
                     bts.ExecutionNode, pipeline_run.root_execution_id
                 )
                 if execution_node:
-                    task_spec = structures.TaskSpec.from_json_dict(
-                        execution_node.task_spec
+                    pipeline_name = database_ops.get_pipeline_name_from_task_spec(
```
morgan-wowk: Mixing migrations and data interactions for business logic in one module is starting to pile up. I'm feeling like we need to:  
  
1\. Split the migration-related code into a separate module from methods that support our business logic in general.  
2\. As a team, work towards a migration solution that has versions so we no longer need to do "checks" on every startup.

morgan-wowk: (1) from the list above would be great to follow up on soon / in a separate PR. (2) is more of a long term initiative we can all work towards.

tests/test_database_ops.py
```diff
@@ -206,203 +278,719 @@ def test_matches_exact_key(
...
+        with session_factory() as session:
+            database_ops._backfill_pipeline_names_from_component_spec(session=session)
+            session.commit()
+
+        with session_factory() as session:
+            annotations = service.list_annotations(session=session, id=run.id)
+        assert annotations[key] == "fallback-name"
+
+
+class TestCreatedByBackfill:
```
morgan-wowk, line 833: Eventually when we move to a versioned migration system, with up/down rollbacks, I'd vote we remove these tests. I have worked with migrations across nearly a dozen apps and wouldn't typically have tests for them. However, because we are literally doing checks and custom-shaped methods, it's good for now.

morgan-wowk (approved):

PR #130, opened 2026-02-26
feat: Search within date range in pipeline run API

yuechao-qin:
### TL;DR

Implemented search within date range in Pipeline Runs.

#### Functional

  - Search by date range in `filter_query`. Example query (not URL encoded for example):
    ```
      /api?filter_query={
        "and": [{
          "time_range": {
            "key": "system/pipeline_run.date.created_at",
            "start_time": "2024-01-01T00:00:00Z",
            "end_time": "2024-02-01T00:00:00Z"
          }
        }]
      }
    ```
  - Valid time ranges:
    - Start time only
    - End time only
    - (Start + End) time
    

#### Other

- Added timezone handling to convert aware datetimes to naive UTC for database compatibility (i.e. date range search will convert API timezone to UTC because DB datetime is only UTC)

### How to test?

```
```

The implementation includes comprehensive tests covering:
- Basic time range filtering with start and end times
- Start-only time ranges (no end time)
- Boundary conditions (inclusive start, exclusive end)
- Timezone offset handling (e.g., `+05:30` converted to UTC)
- Negation with `NOT` operator
- Combination with annotation-based filters
- Nested logical operations (`AND`/`OR`)
- Multiple time ranges in the same query

### Why make this change?

- This change enables users to filter pipeline runs by creation date range.
- The implementation properly handles timezone conversions to ensure consistent behavior across different client timezones while maintaining compatibility with the database's naive UTC storage format.

### DIFF
--- a/cloud_pipelines_backend/filter_query_sql.py
+++ b/cloud_pipelines_backend/filter_query_sql.py
@@ -1,4 +1,5 @@
 import base64
+import datetime
 import json
 import enum
 from typing import Any, Final
@@ -16,6 +17,7 @@ _PIPELINE_RUN_KEY_PREFIX: Final[str] = f"{SYSTEM_KEY_PREFIX}pipeline_run."
 class PipelineRunAnnotationSystemKey(enum.StrEnum):
     CREATED_BY = f"{_PIPELINE_RUN_KEY_PREFIX}created_by"
     PIPELINE_NAME = f"{_PIPELINE_RUN_KEY_PREFIX}name"
+    CREATED_AT = f"{_PIPELINE_RUN_KEY_PREFIX}date.created_at"
 
 
 SYSTEM_KEY_SUPPORTED_PREDICATES: dict[PipelineRunAnnotationSystemKey, set[type]] = {
@@ -29,6 +31,9 @@ SYSTEM_KEY_SUPPORTED_PREDICATES: dict[PipelineRunAnnotationSystemKey, set[type]]
         filter_query_models.ValueEqualsPredicate,
         filter_query_models.ValueContainsPredicate,
         filter_query_models.ValueInPredicate,
+    },
+    PipelineRunAnnotationSystemKey.CREATED_AT: {
+        filter_query_models.TimeRangePredicate,
     },
 }
 
@@ -301,8 +306,9 @@ def _predicate_to_clause(
             return _value_contains_to_clause(predicate=predicate)
         case filter_query_models.ValueInPredicate():
             return _value_in_to_clause(predicate=predicate)
+        case filter_query_models.TimeRangePredicate():
+            return _time_range_to_clause(predicate=predicate)
         case _:
-            # TODO: TimeRangePredicate -- not supported currently, will be supported in the future.
             raise NotImplementedError(
                 f"Predicate type {type(predicate).__name__} is not yet implemented."
             )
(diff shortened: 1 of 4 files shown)


cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -361,3 +369,56 @@ def _value_in_to_clause(
             bts.PipelineRunAnnotation.value.in_(predicate.value_in.values),
         ],
     )
+
+
+# ---------------------------------------------------------------------------
```
yuechao-qin, line 374: Add tests for start_time is None.  Changes from code review.

cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -363,3 +369,59 @@ def _value_in_to_clause(
...
+      UTC-aware datetime
+        datetime(2024, 1, 1, 2, 30, 0, tzinfo=UTC)
+              |
+              v  .replace(tzinfo=None) -- strips timezone label
+      Naive datetime
+        datetime(2024, 1, 1, 2, 30, 0)
+              |
+              v  SQLAlchemy literal_binds -- adds microsecond precision
+      SQL string
+        '2024-01-01 02:30:00.000000'  <-- matches DB storage format
```
Ark-kun, line 411: I'm not sure this will work properly. We take naive timestamp from the user and convert it to certain timezone. I'm not sure the result is correct. We do not know the user's timezone. The backend timezone will be used (which one?).

If the user gives us naive timestamps, let's treat them as UTC.

cloud_pipelines_backend/filter_query_sql.py
```diff
@@ -363,3 +369,59 @@ def _value_in_to_clause(
...
+            "time_range only supports key "
+            f"{PipelineRunAnnotationSystemKey.CREATED_AT!r}, got {tr.key!r}"
+        )
+    # Convert aware datetimes to naive UTC to match DB storage format.
+    clauses: list[sql.ColumnElement] = []
+    if tr.start_time is not None:
+        start_utc = tr.start_time.astimezone(datetime.timezone.utc).replace(tzinfo=None)
+        clauses.append(bts.PipelineRun.created_at >= start_utc)
+    if tr.end_time is not None:
+        end_utc = tr.end_time.astimezone(datetime.timezone.utc).replace(tzinfo=None)
```
Ark-kun, line 425: Just curious: What happens if we do not do `.replace(tzinfo=None)`

Ark-kun (changes requested):
Let's make sure the timezones work well.

cloud_pipelines_backend/filter_query_sql.py
yuechao-qin: I'm confused on this comment. Let me make sure we're aligned:

- My implementation will NOT ALLOW naive timestamps.  It requires timezone-aware timestamps only.  The Pydantic model uses `AwareDatetime` for start/end time paramters.  If the client does not provide a timezone (i.e. naive), Pydantic will return a 422 validation error.
- Thus the user is required to send use timezone-aware timestamps.  We convert it to UTC, since that's what the DB uses.
- Are you suggesting we allow naive timestamps?  I rather not do that, because naive timestamps are ambiguous, we can't know if the user meant UTC, their local timezone, or the server's timezone. Silently assuming UTC could give wrong results for users in other timezones. Requiring an explicit timezone removes all guesswork and makes the API contract unambiguous.

cloud_pipelines_backend/filter_query_sql.py
yuechao-qin: If we did not do .replace(tzinfo=None), it will keep the timezone in the datetime used in the WHERE clause, which does not conform to the DB's naive datetime format. Depending on the DB, it'll react differently. Worst case is the datetime comparisons will be wrong.

cloud_pipelines_backend/filter_query_sql.py
morgan-wowk: I believe it would be more straightforward if we just support UTC - that is probably what Alexey is suggesting.

The comment right now indicates that we support ISO 8601 format in general, which includes a variety of timezone offsets, not just UTC.

I think the solution you have appears like it would work well. It does introduce flexibility that we probably don't need; Yet, I don't imagine there would be a significant difference in lines of code. Just more careful testing.

The alternative would be to strictly enforce the format `2024-01-01T08:00:00Z` on the API rather than having to do the conversion because we support offsets.

Ark-kun: My main concern was with the `.astimezone(utc)`.  
For example, if timestamp is a naive UTC timestamp, then .astimezone(utc) will change it.  
  
\>If the client does not provide a timezone (i.e. naive), Pydantic will return a 422 validation error.  
  
OK. This lifts my concern.

PR #131, opened 2026-02-26
feat: API Server - Pipeline Runs - Changed the pagination token to a cursor

yuechao-qin:
### TL;DR

Closes https://github.com/Shopify/oasis-frontend/issues/547

Replaced offset-based pagination with cursor-based pagination for searching pipeline runs.

#### Old format (offset-based, base64-encoded)

The `page_token` was a base64-encoded JSON blob containing `offset`, `filter`, and `filter_query`:

```
base64({"offset": 10, "filter": null, "filter_query": null})
  = eyJvZmZzZXQiOiAxMCwgImZpbH=
```

API request for page 2:

```
GET /api/pipeline_runs?page_token=eyJvZmZzZXQiOiAxMCwgImZpbH=
```

#### New format (cursor-based, plain text)

The `page_token` is now a plain-text `created_at~id` cursor pointing to the last row of the previous page:

```
2024-02-01T09:00:00+00:00~018d8fff
```

API request for page 2:

```
GET /api/pipeline_runs?page_token=2024-02-01T09:00:00+00:00~018d8fff
```

#### Functional

- **Pagination mechanism**: Switched from offset/limit to cursor-based pagination using `created_at` and `id` fields

#### Other

- **Database optimization**: Added composite index `ix_pr_created_at_desc_id_desc` on `(created_at DESC, id DESC)` for efficient cursor queries
- **Query structure**: Modified SQL queries to use tuple comparison `(created_at, id) < (cursor_created_at, cursor_id)` instead of `OFFSET`
- **API compatibility**: Maintained existing `page_token` parameter while changing internal implementation

### How to test?

```
```

- Verify pagination works correctly with multiple pages of pipeline runs
- Test that cursor pagination maintains stable ordering even when new records are inserted between page requests
- Confirm no duplicate records appear across paginated results
- Validate that invalid page tokens raise appropriate errors
- Test pagination combined with filter queries

### Why make this change?

Cursor-based pagination provides several advantages over offset-based pagination:
- **Performance**: Eliminates expensive `OFFSET` (reads all rows that are skipped) operations on large datasets
- **Consistency**: Prevents duplicate or missing records when data is inserted/deleted during pagination
- **Scalability**: Query performance remains constant regardless of page depth
- **Stability**: Results remain consistent even with concurrent modifications to the dataset

### DIFF
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -208,26 +208,27 @@ class PipelineRunsApiService_Sql:
         include_pipeline_names: bool = False,
         include_execution_stats: bool = False,
     ) -> ListPipelineJobsResponse:
-        where_clauses, offset, next_token = filter_query_sql.build_list_filters(
+        where_clauses = filter_query_sql.build_list_filters(
             filter_value=filter,
             filter_query_value=filter_query,
-            page_token_value=page_token,
+            cursor_value=page_token,
             current_user=current_user,
-            page_size=self._DEFAULT_PAGE_SIZE,
         )
 
         pipeline_runs = list(
             session.scalars(
                 sql.select(bts.PipelineRun)
                 .where(*where_clauses)
-                .order_by(bts.PipelineRun.created_at.desc())
-                .offset(offset)
+                .order_by(
+                    bts.PipelineRun.created_at.desc(),
+                    bts.PipelineRun.id.desc(),
+                )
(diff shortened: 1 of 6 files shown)


Ark-kun:
One small advantage of opaque next_page_token is that the filter and cursor/offset are tied together.

Ark-kun:
We can decode the created_at from the id. Do we need it in the cursor?

Ark-kun (approved):

PR #135, opened 2026-02-28
feat: Add `include_sql` to Search Pipeline Run API

yuechao-qin:
### TL;DR

Closes https://github.com/Shopify/oasis-frontend/issues/548

Added optional SQL query debugging to the pipeline runs list API endpoint.

#### Functional

- Added `include_sql` parameter to API

```
GET /api/pipeline_runs/?include_sql=true&filter_query=<QUERY>
```

- Added `sql` field to `ListPipelineJobsResponse` to optionally return the generated SQL query

```
{
  "pipeline_runs": [...],
  "next_page_token": <TOKEN>,
  "sql": "SELECT * FROM pipeline_run WHERE ... ORDER BY ..."
}
```

### How to test?

```
uv run pytest tests/test_api_server_sql.py
```

- Call the list endpoint with `include_sql=True` to verify the SQL query is returned in the response
- Test with various filter queries and pagination tokens to ensure the SQL compilation works correctly
- Verify that when `include_sql=False` or is omitted, the `sql` field remains `None`

### Why make this change?

- This enables debugging and monitoring of the actual SQL queries generated by the pipeline runs list endpoint.
- This is required by E2E testing (PR [here](https://app.graphite.com/github/pr/Shopify/oasis-backend/116/test-Local-E2E-tests-for-Search-Pipeline-Run-API)) to be able to log and report what SQL statement was used per API request.

### DIFF
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -84,6 +84,7 @@ class GetPipelineRunResponse(PipelineRunResponse):
 class ListPipelineJobsResponse:
     pipeline_runs: list[PipelineRunResponse]
     next_page_token: str | None = None
+    sql: str | None = None
 
 
 class PipelineRunsApiService_Sql:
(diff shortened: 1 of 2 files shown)


Ark-kun:
I'm not sure we should support this in public API.
This could also become problematic if we support alternative non-SQL search methods like ElasticSearch/OpenSearch.

It's pretty simple to locally compile arbitrary filter_query or page_token to arbitrary SQL dialect.

Let's discuss the primary use case and decide how to accomplish that goal without adding this debug information to the public API surface.

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -68,6 +68,7 @@ class GetPipelineRunResponse(PipelineRunResponse):
 class ListPipelineJobsResponse:
     pipeline_runs: list[PipelineRunResponse]
     next_page_token: str | None = None
+    sql: str | None = None
```
yuechao-qin, line 87: Brainstorming ideas:
- add "_" in front of paramter "_include_sql"
- hide include_sql from openapi
- add this to oasis-backend not tangle (https://oasis.shopify.io/api/stats/quick?format=html)

PR #136, opened 2026-03-04
chore: Reduce RuntimeErrors for expected states with no container execution

morgan-wowk:
### TL;DR

Improved error handling for container executions by defining expected statuses where container execution records may be missing and updating error logic accordingly.

### What changed?

Added a new constant `CONTAINER_STATUSES_WITH_MISSING_EXECUTION_EXPECTED` that defines container execution statuses where missing execution records are expected (INVALID, UNINITIALIZED, QUEUED, WAITING_FOR_UPSTREAM, SYSTEM_ERROR, CANCELLED, SKIPPED). Updated `get_container_execution_state()` to throw `ItemNotFoundError` instead of `RuntimeError` when container execution is missing for expected statuses. Modified `get_container_execution_log()` to use the new status validation logic and throw `ItemNotFoundError` for expected missing states while maintaining special handling for SYSTEM_ERROR status.

### How to test?

Test API endpoints for container executions in various states:
- Call `get_container_execution_state()` with executions in QUEUED, CANCELLED, or SKIPPED status and verify it returns `ItemNotFoundError` with appropriate messaging
- Call `get_container_execution_log()` with executions in expected missing states and verify it returns `ItemNotFoundError` for most statuses or error response for SYSTEM_ERROR
- Verify that executions in unexpected states still throw `RuntimeError` when container execution records are missing

### Why make this change?

This change provides more appropriate error handling by distinguishing between expected and unexpected cases where container execution records are missing. For certain statuses like QUEUED or CANCELLED, it's normal for the container execution record to not exist yet or ever, so the API should handle these cases gracefully rather than treating them as unexpected runtime errors.

### DIFF
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -679,8 +679,14 @@ class ExecutionNodesApiService_Sql:
             raise ItemNotFoundError(f"Execution with {id=} does not exist.")
         container_execution = execution.container_execution
         if not container_execution:
-            raise RuntimeError(
-                f"Execution with {id=} does not have container execution information."
+            status = execution.container_execution_status
+            if not status or status not in bts.CONTAINER_STATUSES_WITH_MISSING_EXECUTION_EXPECTED:
+                raise RuntimeError(
+                    f"Execution with {id=} has status {status} "
+                    f"but is missing its container execution record."
+                )
+            raise errors.ItemNotFoundError(
+                f"Execution with {id=} has status {status.value} and does not have container execution information. This is expected for this status."
             )
         return GetContainerExecutionStateResponse(
             status=container_execution.status,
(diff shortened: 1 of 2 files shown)


cloud_pipelines_backend/api_server_sql.py
```diff
@@ -679,8 +679,14 @@ def get_container_execution_state(
             raise ItemNotFoundError(f"Execution with {id=} does not exist.")
         container_execution = execution.container_execution
         if not container_execution:
```
morgan-wowk, line 681: **GET /container_state** **endpoint** **enters** **this** **codepath**

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -756,16 +762,19 @@ def get_container_execution_log(
             system_error_exception_full or orchestration_error_message
         )
         if not container_execution:
```
morgan-wowk, line 764: GET /**container_execution_log** **enters** **this** **codepath**

cloud_pipelines_backend/api_server_sql.py
morgan-wowk: No reason to return a 200 response if there is a SYSTEM_ERROR in this case.

cloud_pipelines_backend/api_server_sql.py
morgan-wowk: Even if we don't have a container execution, we can still have useful SYSTEM_ERROR information. So when there's a SYSTEM_ERROR, we use GetContainerExecutionLogResponse. Otherwise, we use ItemNotFoundError.

PR #137, opened 2026-03-04
feat: Propagate kubernetes launcher errors to API

morgan-wowk:
### TL;DR

Enhanced Kubernetes error handling to extract and display meaningful error messages from API exceptions instead of dumping raw pod/job specifications.

### What changed?

Added specific handling for `k8s_client_lib.ApiException` in both pod and job creation methods within the Kubernetes launchers. The new error handling extracts human-readable messages from the Kubernetes API response body and stores orchestration error messages in execution extra data for better debugging visibility.

### How to test?

1. Trigger a Kubernetes pod or job creation that would fail due to admission webhook rejection or other API validation errors
2. Verify that the error message displays the actual Kubernetes API error reason instead of a serialized pod/job specification
3. Check that the orchestration error message is stored in the execution's extra data for system errors

### Why make this change?

Kubernetes API returns structured JSON error responses with actionable error messages (like admission webhook rejection reasons), but the previous implementation would dump the entire pod/job specification instead. This change provides much more useful debugging information by surfacing the actual root cause of failures.

### DIFF
--- a/cloud_pipelines_backend/launchers/kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/kubernetes_launchers.py
@@ -501,6 +501,11 @@ class _KubernetesPodLauncher(
                 body=pod,
                 _request_timeout=self._request_timeout,
             )
+        except k8s_client_lib.ApiException as ex:
+            k8s_message = _extract_kubernetes_error_message(ex)
+            raise interfaces.LauncherError(
+                k8s_message or f"Failed to create pod: Kubernetes API error ({ex.status} {ex.reason})"
+            ) from ex
         except Exception as ex:
             raise interfaces.LauncherError(
                 f"Failed to create pod: {_kubernetes_serialize(pod)}"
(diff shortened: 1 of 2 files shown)


cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -586,6 +586,11 @@ def generate_execution_log_uri(
                     bts.ContainerExecutionStatus.SYSTEM_ERROR
                 )
                 record_system_error_exception(execution=execution, exception=ex)
+                if execution.extra_data is None:
```
Ark-kun, line 589: Hmm. Why do you want to put this information in the another field? The error message (brief, full) is already put in extra_data.
P.S. There is `_record_orchestration_error_message` function that does slightly more.

cloud_pipelines_backend/launchers/kubernetes_launchers.py
```diff
@@ -501,6 +501,11 @@ def launch_container_task(
                 body=pod,
                 _request_timeout=self._request_timeout,
             )
+        except k8s_client_lib.ApiException as ex:
+            k8s_message = _extract_kubernetes_error_message(ex)
+            raise interfaces.LauncherError(
+                k8s_message or f"Failed to create pod: Kubernetes API error ({ex.status} {ex.reason})"
```
Ark-kun, line 507: Should we include the Pod that we were trying to create (as the current message does)? I think it's very important for debugging pod creation issues.

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: This was meant to reach tangle-ui so that we can have a clean, shortened JSON object that can be used to display a GPU Quota Group element. The team went in another direction (which I like) and this is not needed. Also, yes, it was pretty redundant

PR #147, opened 2026-03-10
fix: PipelineRun Search - Fixed backfill for MySQL when annotation value length > 255

yuechao-qin:
### TL;DR

Added string truncation to prevent database errors when `pipeline names` or `created by` exceed the 255-character VARCHAR limit in annotation `value` column.  This fix is applied during:

- Backfilling pipeline name to annotations table.
- Pipeline create/start and the `created by` or `pipeline name` > 255 characters:
  - Truncation will happen ONLY for annotations table, run will be allowed to start with a warning log.

### What changed?

- Added `_truncate_for_annotation()` helper function that truncates strings to 255 characters and logs warnings when truncation occurs
- Applied truncation to `created_by` and `pipeline_name` values in `_mirror_system_annotations()` before inserting into the database
- Updated database backfill operations to truncate pipeline names using `sqlalchemy.func.substr()` 
- Extracted the VARCHAR length limit (255) into a shared constant `_STR_MAX_LENGTH`
- Added comprehensive test coverage including a custom SQLite fixture that mimics MySQL's VARCHAR overflow behavior

### How to test?

```
uv run pytest tests/test_api_server_sql.py tests/test_database_ops.py
```

Run the new test classes:
- `TestTruncateForAnnotation` - Unit tests for the truncation helper function
- `TestAnnotationValueOverflow` - Integration tests that verify truncation prevents database errors
- Extended `TestBackfillPipelineNames` - Tests that backfill operations properly truncate long values

The tests use a special SQLite fixture with triggers that simulate MySQL's strict VARCHAR enforcement.

### Why make this change?

This is blocking production from being released because the backfill needs to happen first before release is successful.

### DIFF
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -1226,6 +1226,29 @@ _ArtifactNodeOrDynamicDataType = typing.Union[
 ]
 
 
+def _truncate_for_annotation(
+    *,
+    value: str,
+    field_name: str,
+    pipeline_run_id: bts.IdType,
+) -> str:
+    """Truncate a string to fit the annotation VARCHAR column.
+
+    Returns the value unchanged if it fits within _STR_MAX_LENGTH,
+    otherwise truncates and logs a warning with the run ID and
+    original length.
+    """
+    max_len = bts._STR_MAX_LENGTH
+    if len(value) <= max_len:
+        return value
+
+    _logger.warning(
+        f"Truncating {field_name} annotation for run {pipeline_run_id}: "
+        f"{len(value)} chars -> {max_len} chars"
+    )
+    return value[:max_len]
+
+
 def _mirror_system_annotations(
     *,
     session: orm.Session,
@@ -1233,8 +1256,20 @@ def _mirror_system_annotations(
     created_by: str | None,
     pipeline_name: str | None,
 ) -> None:
-    """Mirror pipeline run fields as system annotations for filter_query search."""
+    """Mirror pipeline run fields as system annotations for filter_query search"""
+
+    # TODO: The original pipeline_run.created_by and the pipeline name stored in
+    # extra_data / task_spec are saved untruncated, while the annotation mirror
+    # is truncated to VARCHAR(255). This creates a data parity mismatch between
+    # the source columns and their annotation copies. Revisit this to either
+    # widen the annotation column or enforce the same limit at the source.
+
     if created_by:
+        created_by = _truncate_for_annotation(
+            value=created_by,
+            field_name=filter_query_sql.PipelineRunAnnotationSystemKey.CREATED_BY,
+            pipeline_run_id=pipeline_run_id,
+        )
         session.add(
             bts.PipelineRunAnnotation(
                 pipeline_run_id=pipeline_run_id,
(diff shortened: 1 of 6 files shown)


cloud_pipelines_backend/backend_types_sql.py
```diff
@@ -117,7 +120,7 @@ class _TableBase(orm.MappedAsDataclass, orm.DeclarativeBase, kw_only=True):
         # structures.TypeSpecType: sql.JSON,  # !!! Gotcha. This causes str map to JSON
         # SqlIOTypeStruct: sql.JSON,
         # PipelineSpec: sql.JSON,
-        str: sql.String(255),
+        str: sql.String(_STR_MAX_LENGTH),
```
morgan-wowk, line 123: praise: I'm glad you made this a constant

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -1226,15 +1226,51 @@ def list_secrets(
...
+    original length.
+    """
+    max_len = bts._STR_MAX_LENGTH
+    if len(value) <= max_len:
+        return value
+
+    _logger.warning(
+        f"Truncating {field_name} annotation for run {pipeline_run_id}: "
+        f"{len(value)} chars -> {max_len} chars. "
+        f"Original value (first 80 chars): {value[:80]!r}"
```
morgan-wowk: I'm not sure the original value here provides much value since it's less characters than the portion of the name we do persist.

morgan-wowk: :resolvable / you decide

yuechao-qin: I agree, I was torn whether we should add it or not, your comment adds more weight where I should lean.

Removing printing the value.

tests/test_api_server_sql.py
```diff
@@ -582,6 +582,130 @@ def test_delete_annotation_rejects_system_key(self, session_factory, service):
...
+    """Reproduction tests using mysql_varchar_limit_session_factory (SQLite TRIGGER
+    enforcement). These tests prove that >255 char values are rejected,
+    mimicking MySQL's DataError 1406.
+
+    Covers all write paths into pipeline_run_annotation:
+    - set_annotation(): long key, long value
+    - create() via _mirror_system_annotations(): long pipeline_name, long created_by
+    """
+
+    def test_set_annotation_long_value_raises_on_overflow(
```
morgan-wowk, line 628: This test feels highly simulated. In [conftest.py](http://conftest.py) we simulate an upper boundary on data length. Then, we have a test which tests the simulated state.

It seems to prove that there are scenarios in our `_create_run` codepath where an error can still occur, contrary to our goals of truncating in order to avoid errors.

morgan-wowk: Should we be truncating the value in this case rather than letting it produce an error?

tests/test_api_server_sql.py
```diff
@@ -582,6 +582,130 @@ def test_delete_annotation_rejects_system_key(self, session_factory, service):
...
+            ):
+                service.set_annotation(
+                    session=session,
+                    id=run.id,
+                    key="team",
+                    value="v" * 300,
+                    user_name="user1",
+                )
+
+    def test_set_annotation_long_key_raises_on_overflow(
```
morgan-wowk, line 653: Same feedback: https://app.graphite.com/github/pr/TangleML/tangle/147?ref=gt-pasteable-stack#comment-PRRC_kwDOOC2o5s6tYaSo

tests/test_api_server_sql.py
morgan-wowk: I see now that you are (1) creating a run successfully, and (2), then using that run to set an annotation, and expecting `set_annotation` to fail.

morgan-wowk: Received a response in Slack:

```
I meant to add a comment that the tests is showing there is an issue with set_annotation but didn't want to convolute this PR with that fix.  Will require another PR to fix the issue.
```

tests/conftest.py
Ark-kun, line 1: We should avoid globally-defined fixtures that are only used in some specific tests.

PR #166, opened 2026-03-13
fix: Robust and Idempotent Backfill for Search Pipeline Run API

yuechao-qin:
### TL;DR

Refactored database migration logic to ensure data parity by creating system annotations for all pipeline runs, even when source values are null or empty, and moved backfill functions to a dedicated module with comprehensive error handling.

### What changed?

- **System annotation mirroring now ensures data parity**: The `_mirror_system_annotations` function always creates annotation rows for `created_by` and `pipeline_name`, storing empty string `""` when source values are null/empty, with warning logs for null cases
- **New dedicated migration module**: Created `database_migrate.py` containing three idempotent backfill functions:
  - `backfill_created_by_annotations`: Uses `COALESCE` to handle null values
  - `backfill_pipeline_names_from_extra_data`: Extracts from JSON with null filtering  
  - `backfill_pipeline_names_from_component_spec`: Extracts from nested JSON path with anti-join logic
- **Robust orchestration**: `run_all_annotation_backfills` wraps all backfills in try-catch with configurable skip guards and single transaction commit
- **Enhanced test coverage**: Added 1600+ lines of comprehensive tests covering idempotency, order independence, data parity, error handling, and edge cases
- **Updated existing tests**: Modified assertions to expect empty string annotations instead of missing keys

### How to test?

Run the existing test suite - the new `test_database_migrate.py` provides extensive coverage including:
- Basic backfill functionality for both sources
- Idempotency verification (safe to run multiple times)
- Order independence between pipeline name sources
- Data parity validation (every run gets annotations)
- Error handling and transaction rollback scenarios
- Truncation behavior for long values

### Why make this change?

- **Data consistency**: Ensures every pipeline run has system annotations for reliable filtering/querying, eliminating gaps where some runs lack annotation rows
- **Improved maintainability**: Separates migration logic from general database operations with better organization and comprehensive documentation
- **Production reliability**: Adds proper error handling so migration failures don't block application startup, with detailed logging for debugging
- **Database portability**: Uses SQLAlchemy abstractions for cross-database compatibility (SQLite, MySQL, PostgreSQL)


cloud_pipelines_backend/database_migrate.py
```diff
@@ -0,0 +1,542 @@
...
+                session=session,
+                auto_commit=False,
+            )
+            backfill_pipeline_names_from_component_spec(
+                session=session,
+                auto_commit=False,
+            )
+            # TODO: Do we need a final catchall backfill that inserts empty string
+            # for all pipeline names, which happens to not have a name in
+            # component_spec nor extra_data?
```
yuechao-qin: Do we need this?  A catchall backfill for `pipeline name` if 1) `extra_data` and 2) `component_spec_name` doesn't exist/had issues?

cloud_pipelines_backend/database_migrate.py
```diff
@@ -0,0 +1,542 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+def _is_pipeline_run_annotation_key_already_backfilled(
+    *,
+    session: orm.Session,
+    key: str,
+) -> bool:
+    """Return True if at least one annotation with the given key exists."""
```
Ark-kun: This is not safe anymore.
Due to changes in API Server, pipeline names are already being inserted into the DB (while bulk inserts weren't added).

The best way to check whether the backfill is complete is to compare the number of rows in pipeline runs table and the number of pipeline run annotations.

cloud_pipelines_backend/database_migrations.py
```diff
@@ -0,0 +1,542 @@
...
+    ].as_string()
+    # Step 4: SUBSTR truncation for VARCHAR(255)
+    truncated_name = sqlalchemy.func.substr(
+        name_expr,
+        1,
+        bts._STR_MAX_LENGTH,
+    )
+    existing_ann = orm.aliased(bts.PipelineRunAnnotation)
+
+    stmt = sqlalchemy.insert(bts.PipelineRunAnnotation).from_select(
```
Ark-kun, line 585: JFYI: If we expect most annotations to be missing, we could use the `INSERT ... ON CONFLICT DO NOTHING` statement: https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#sqlalchemy.dialects.sqlite.Insert.on_conflict_do_nothing 
But if most annotations exist, then outer join might be better.

cloud_pipelines_backend/database_migrations.py
```diff
@@ -0,0 +1,542 @@
...
+    if auto_commit:
+        session.commit()
+    return rowcount
+
+
+def run_all_annotation_backfills(
+    *,
+    session: orm.Session,
+    do_skip_already_backfilled: bool,
+) -> None:
```
yuechao-qin: Alexey: Look into different transaction modes 

https://www.postgresql.org/docs/current/transaction-iso.html

For backfills let's use the most reliable mode (serializable).

For example: expectations is if there are multiple migrations, that failure shouldn't of been duplicate entry, but a transaction conflict error (i.e. notice rows being added from another migration).

Try to see if this can be simulated/reproduced locally.

cloud_pipelines_backend/database_migrate.py
yuechao-qin: Update backfill from component spec, that if name does not exist to backfill with empty string.

Ark-kun (approved):

cloud_pipelines_backend/database_migrations.py
Ark-kun, line 1: Style: Modules should be names as plural or uncountable nouns.
Example: `database_migrations`

cloud_pipelines_backend/database_migrations.py
yuechao-qin: This seems really neat!

However, looking at it, it's only for SQLite dialect.

cloud_pipelines_backend/database_migrations.py
yuechao-qin: > UPDATE: Talked with @Ark-kun will test and use this in future PR.  Different behaviors do not require different DB dialects AND behaviors are not intrusive, so more pro vs. con to going with this approach.

I looked into this, unfortunately it's not portable across all SQL DBs, it has different behaviors.

## The Portability Problem

This is the dealbreaker:

| DB | Serializable behavior for `INSERT ... FROM SELECT` race |
|---|---|
| **PostgreSQL** | Clean serialization failure (SSI predicate locks detect read→write dependency) |
| **MySQL** | **Still throws IntegrityError** — InnoDB's gap locks don't fully prevent this pattern for `INSERT ... SELECT` |
| **SQLite** | **Blocks** (file-level lock) — no error at all, just waits |

yuechao-qin: https://app.graphite.com/github/pr/TangleML/tangle/198

PR #167, opened 2026-03-13
Add optional total_count field to ListPipelineJobsResponse

Mbeaulne:
### TL;DR

Added `total_count` field to the `ListPipelineJobsResponse` to provide the total number of pipeline runs matching the query criteria when requested.

### What changed?

- Added `total_count: int | None = None` field to `ListPipelineJobsResponse` class
- Added `include_total_count: bool = False` parameter to the `list` method in `PipelineRunsApiService_Sql`
- When `include_total_count` is True, the method executes a count query using `sql.func.count()` with the same filter conditions as the main query
- The `total_count` is populated in the response object when the parameter is enabled

### How to test?

Run the existing test suite which now includes comprehensive tests for the `total_count` functionality:
- `test_list_total_count_not_included_by_default` - verifies count is None by default
- `test_list_total_count_empty` - verifies count is 0 when no runs exist
- `test_list_total_count_matches_results` - confirms count matches actual results
- `test_list_total_count_with_pagination` - ensures count remains consistent across paginated requests
- `test_list_total_count_with_filter` - validates count respects filter conditions

### Why make this change?

This enhancement allows clients to optionally retrieve the total number of pipeline runs that match their query criteria, which is essential for implementing proper pagination controls and displaying accurate result counts in user interfaces. The feature is opt-in to avoid performance overhead when the total count is not needed.

### DIFF
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -84,6 +84,7 @@ class GetPipelineRunResponse(PipelineRunResponse):
 class ListPipelineJobsResponse:
     pipeline_runs: list[PipelineRunResponse]
     next_page_token: str | None = None
+    total_count: int | None = None
 
 
 class PipelineRunsApiService_Sql:
@@ -206,6 +207,7 @@ class PipelineRunsApiService_Sql:
         current_user: str | None = None,
         include_pipeline_names: bool = False,
         include_execution_stats: bool = False,
+        include_total_count: bool = False,
     ) -> ListPipelineJobsResponse:
         where_clauses, offset, next_token = filter_query_sql.build_list_filters(
             filter_value=filter,
@@ -225,6 +227,14 @@ class PipelineRunsApiService_Sql:
             ).all()
         )
 
+        total_count = None
+        if include_total_count:
+            total_count = session.scalar(
+                sql.select(sql.func.count())
+                .select_from(bts.PipelineRun)
+                .where(*where_clauses)
+            )
+
         next_page_token = (
             next_token if len(pipeline_runs) >= self._DEFAULT_PAGE_SIZE else None
         )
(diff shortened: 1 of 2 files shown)


cloud_pipelines_backend/api_server_sql.py
```diff
@@ -225,6 +227,14 @@ def list(
             ).all()
         )
 
+        total_count = None
+        if include_total_count:
+            total_count = session.scalar(
+                sql.select(sql.func.count())
+                .select_from(bts.PipelineRun)
+                .where(*where_clauses)
```
morgan-wowk, line 235: At a glance this looks like it will only return you the count for the current page. Can you confirm?

morgan-wowk: Although, your tests seem to suggest otherwise

morgan-wowk: Oh I see what's happening. This is based on limit + offset. Which will soon change:  
  
https://app.graphite.com/github/pr/TangleML/tangle/131?ref=gt-pasteable-stack  
  
I guess, let's adjust this solution accordingly

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -225,6 +227,14 @@ def list(
             ).all()
         )
 
+        total_count = None
```
Ark-kun, line 230: This looks problematic as this makes queries about 4000+x slower. Basically, we:  
1) Run same query twice  
2) The 2nd time we run it without any limits.  
This is really bad for query performance.  
As the DB grows, these queries become slower and slower. For some heavy queries like substring match (search by name), the slowdown will be immense.  
Normal queries search for "First 10 matches", scanning the DB until 10 matches are found. But the new "total_count" query always scans the whiole DB (40000+ runs).

morgan-wowk (changes requested):
Some chat in Slack about performance at scale

https://shopify.slack.com/archives/C08D1B6V32N/p1773475475243729?thread_ts=1773429428.873269&cid=C08D1B6V32N

PR #174, opened 2026-03-18
feat: Measure execution status duration

morgan-wowk:
### **Changes**

- Adds `status_updated_at` column to `ExecutionNode` table to track when execution status last changed
- Implements SQLAlchemy event listener to automatically update `status_updated_at` timestamp when `container_execution_status` changes
- Creates `execution_status_transition_duration` histogram metric to measure time spent in each execution status
- Adds `_transition_execution_status()` helper function to centralize status updates and metric recording across all status transitions
- Implements database migration logic to add the new `status_updated_at` column to existing tables
- Replaces direct status assignments throughout the orchestrator with calls to the new transition helper function

##### **Note: Attribute** **names** **have** **since** **changed to** `execution_status_` prefix

Local smoke-test and verification completed :white_check_mark:

### DIFF
--- a/cloud_pipelines_backend/backend_types_sql.py
+++ b/cloud_pipelines_backend/backend_types_sql.py
@@ -424,7 +424,10 @@ class ExecutionNode(_TableBase):
         repr=False,
     )
 
+    _status_changed: bool = dataclasses.field(default=False, init=False, repr=False)
 
+
+EXECUTION_NODE_EXTRA_DATA_STATUS_HISTORY_KEY = "container_execution_status_history"
 EXECUTION_NODE_EXTRA_DATA_SYSTEM_ERROR_EXCEPTION_MESSAGE_KEY = (
     "system_error_exception_message"
 )
--- a/cloud_pipelines_backend/instrumentation/metrics.py
+++ b/cloud_pipelines_backend/instrumentation/metrics.py
@@ -22,7 +22,25 @@ Bad example:
 - Instrument: orchestrator_execution_system_errors
 """
 
+import datetime
+import enum
+import logging
+
 from opentelemetry import metrics as otel_metrics
+from sqlalchemy import event as sql_event
+from sqlalchemy import orm
+
+from .. import backend_types_sql
+
+_logger = logging.getLogger(__name__)
+
+
+class MetricUnit(str, enum.Enum):
+    """UCUM-style unit strings accepted by the OTel SDK."""
+
+    SECONDS = "s"
+    ERRORS = "{error}"
+
 
 # ---------------------------------------------------------------------------
 # tangle.orchestrator
@@ -32,5 +50,47 @@ orchestrator_meter = otel_metrics.get_meter("tangle.orchestrator")
 execution_system_errors = orchestrator_meter.create_counter(
     name="execution.system_errors",
     description="Number of execution nodes that ended in SYSTEM_ERROR status",
-    unit="{error}",
+    unit=MetricUnit.ERRORS,
 )
+
+execution_status_transition_duration = orchestrator_meter.create_histogram(
+    name="execution.status_transition.duration",
+    description="Duration an execution spent in a status before transitioning to the next status",
+    unit=MetricUnit.SECONDS,
+)
+
+
+# ---------------------------------------------------------------------------
+# SQLAlchemy event listeners
+# ---------------------------------------------------------------------------
+
+_HISTORY_KEY = backend_types_sql.EXECUTION_NODE_EXTRA_DATA_STATUS_HISTORY_KEY
+
+
+@sql_event.listens_for(orm.Session, "before_commit")
+def _handle_before_commit(session: orm.Session) -> None:
+    for obj in list(session.new) + list(session.dirty):
+        if not isinstance(obj, backend_types_sql.ExecutionNode):
+            continue
+        if not obj._status_changed:
+            continue
+        history: list = (obj.extra_data or {}).get(_HISTORY_KEY, [])
+        if len(history) >= 2:
+            prev = history[-2]
+            curr = history[-1]
+            prev_time = datetime.datetime.fromisoformat(prev["first_observed_at"])
+            curr_time = datetime.datetime.fromisoformat(curr["first_observed_at"])
+            try:
+                execution_status_transition_duration.record(
+                    (curr_time - prev_time).total_seconds(),
+                    attributes={
+                        "execution.status.from": prev["status"],
+                        "execution.status.to": curr["status"],
+                    },
+                )
+            except Exception:
+                _logger.warning(
+                    f"Failed to record status transition metric for execution {obj.id!r}",
+                    exc_info=True,
+                )
+        obj._status_changed = False
--- a/cloud_pipelines_backend/orchestrator_sql.py
+++ b/cloud_pipelines_backend/orchestrator_sql.py
@@ -9,6 +9,7 @@ from typing import Any
 
 
 import sqlalchemy as sql
+from sqlalchemy import event as sql_event
 from sqlalchemy import orm
 
 from cloud_pipelines.orchestration.storage_providers import (
@@ -1101,3 +1102,35 @@ def _maybe_get_small_artifact_value(
             return text
         except Exception:
             pass
+
+
+# ---------------------------------------------------------------------------
+# SQLAlchemy event listeners
+# ---------------------------------------------------------------------------
+
+_HISTORY_KEY = bts.EXECUTION_NODE_EXTRA_DATA_STATUS_HISTORY_KEY
+
+
+@sql_event.listens_for(bts.ExecutionNode.container_execution_status, "set")
+def _handle_container_execution_status_set(
+    execution: bts.ExecutionNode,
+    value: bts.ContainerExecutionStatus | None,
+    _old_value: object,
+    _initiator: object,
+) -> None:
+    if value is None:
+        return
+    if execution.extra_data is None:
+        execution.extra_data = {}
+    history: list = execution.extra_data.get(_HISTORY_KEY, [])
+    if history and history[-1]["status"] == value.value:
+        return
+    entry = {
+        "status": value.value,
+        "first_observed_at": datetime.datetime.now(datetime.timezone.utc).isoformat(),
+    }
+    execution.extra_data = {
+        **execution.extra_data,
+        _HISTORY_KEY: history + [entry],
+    }
+    execution._status_changed = True
--- a/tests/test_sql_event_listeners.py
+++ b/tests/test_sql_event_listeners.py
@@ -1,0 +1,69 @@
+"""Tests for SQLAlchemy event listeners in orchestrator_sql and instrumentation.metrics."""
+
+import unittest.mock
+
+import pytest
+from sqlalchemy import orm
+
+from cloud_pipelines_backend import backend_types_sql as bts
+from cloud_pipelines_backend import database_ops
+from cloud_pipelines_backend import (
+    orchestrator_sql,
+)  # noqa: F401 — registers set listener
+from cloud_pipelines_backend.instrumentation import (
+    metrics,
+)  # noqa: F401 — registers before_commit listener
+
+
+@pytest.fixture()
+def session() -> orm.Session:
+    db_engine = database_ops.create_db_engine(database_uri="sqlite://")
+    bts._TableBase.metadata.create_all(db_engine)
+    with orm.Session(db_engine) as s:
+        yield s
+
+
+class TestStatusHistoryListeners:
+    def test_status_change_appends_history_to_extra_data(
+        self, session: orm.Session
+    ) -> None:
+        node = bts.ExecutionNode(task_spec={})
+        session.add(node)
+        node.container_execution_status = bts.ContainerExecutionStatus.QUEUED
+        session.commit()
+
+        history = node.extra_data[bts.EXECUTION_NODE_EXTRA_DATA_STATUS_HISTORY_KEY]
+        assert len(history) == 1
+        assert history[0]["status"] == bts.ContainerExecutionStatus.QUEUED
+
+    def test_duplicate_status_is_not_appended_to_history(
+        self, session: orm.Session
+    ) -> None:
+        node = bts.ExecutionNode(task_spec={})
+        session.add(node)
+        node.container_execution_status = bts.ContainerExecutionStatus.QUEUED
+        node.container_execution_status = bts.ContainerExecutionStatus.QUEUED
+        session.commit()
+
+        history = node.extra_data[bts.EXECUTION_NODE_EXTRA_DATA_STATUS_HISTORY_KEY]
+        assert len(history) == 1
+
+    def test_second_status_change_records_duration_metric(
+        self, session: orm.Session
+    ) -> None:
+        node = bts.ExecutionNode(task_spec={})
+        session.add(node)
+        node.container_execution_status = bts.ContainerExecutionStatus.QUEUED
+        session.commit()
+
+        node.container_execution_status = bts.ContainerExecutionStatus.RUNNING
+        with unittest.mock.patch.object(
+            metrics.execution_status_transition_duration, "record"
+        ) as mock_record:
+            session.commit()
+
+        mock_record.assert_called_once()
+        assert mock_record.call_args.kwargs["attributes"] == {
+            "execution.status.from": bts.ContainerExecutionStatus.QUEUED,
+            "execution.status.to": bts.ContainerExecutionStatus.RUNNING,
+        }


morgan-wowk:
* **#186** <a href="https://app.graphite.com/github/pr/TangleML/tangle/186?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#174** <a href="https://app.graphite.com/github/pr/TangleML/tangle/174?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/174?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

morgan-wowk:
### Load Impact: `feat: Measure execution status duration`

The commit adds zero additional database calls during steady-state operation. `status_updated_at` is written as part of the existing `UPDATE` statement that already commits `container_execution_status` — SQLAlchemy batches both columns into one flush. Reading the previous status and timestamp before each transition is always an in-memory access from the already-loaded SQLAlchemy instance. The only one-time database cost is a single `SELECT` on every process restart to check for the new column, plus a single `ALTER TABLE` on first deployment. On the metrics side, `record_status_transition` appends to an in-memory histogram bucket and returns immediately; the OTel SDK exports observations to the collector in background batches, adding one additional metric stream with no synchronous I/O on the hot path.

| Area | Impact |
| --- | --- |
| DB calls at startup | \+1 `SELECT` per restart; +1 `ALTER TABLE` on first deploy only |
| DB calls per transition | **None** — `status_updated_at` piggybacks on the existing `UPDATE` |
| DB calls for reading previous status | **None** — in-memory reads from the live SQLAlchemy instance |
| CPU per transition | Negligible — one `datetime.now()` + one histogram bucket increment |
| Memory | Negligible — one additional `datetime` field per loaded `ExecutionNode` |
| Network (OTel export) | One additional histogram metric stream, exported in background batches |

morgan-wowk:
There are two dimensions we can choose to keep, or exclude measuring to begin with:

- RUNNING > SUCCEEDED (some value)
- WAITING_FOR_UPSTREAM > QUEUED (not much value)

They can provide value depending on the use case. However, they can also provide noise. Some observability platforms support filtering. I worry that not all would, and lean towards excluding measurement. We would be able to add separate metrics for total run time in the future if desired.

cloud_pipelines_backend/backend_types_sql.py
```diff
@@ -425,6 +427,20 @@ class ExecutionNode(_TableBase):
     )
 
 
+@sql.event.listens_for(ExecutionNode.container_execution_status, "set")
+def _stamp_execution_status_updated_at(
+    target: ExecutionNode, value, oldvalue, initiator
+) -> None:
```
yuechao-qin: For unused parameters, prefix it with an underscore (e.g. value, _old_value, _initiator)

Also, variables should be snake case.

Add type hinting to the parameters too.

cloud_pipelines_backend/database_ops.py
```diff
@@ -57,7 +56,30 @@ def create_db_engine(
...
+    for table_name, column_name, column_def in _COLUMN_MIGRATIONS:
+        existing_columns = {col["name"] for col in inspector.get_columns(table_name)}
+        if column_name not in existing_columns:
+            with db_engine.connect() as conn:
+                conn.execute(
+                    sqlalchemy.text(
+                        f"ALTER TABLE {table_name} ADD COLUMN {column_name} {column_def}"
+                    )
+                )
+                conn.commit()
```
yuechao-qin: Add * in the function for `keyword-only` arguments, good style even with just 1 parameter.

cloud_pipelines_backend/database_ops.py
```diff
@@ -57,7 +56,30 @@ def create_db_engine(
     return db_engine
 
 
+def _add_columns_if_missing(db_engine: sqlalchemy.Engine) -> None:
+    """Add new nullable columns to existing tables via ALTER TABLE when they are
+    not yet present.  SQLAlchemy's create_all() only creates missing tables, not
+    missing columns, so new columns require an explicit migration step."""
+    _COLUMN_MIGRATIONS: list[tuple[str, str, str]] = [
+        # (table_name, column_name, column_definition)
+        ("execution_node", "status_updated_at", "DATETIME"),
```
yuechao-qin: Let's derive these instead of hard coding them

```
_COLUMN_MIGRATIONS = [
    bts.ExecutionNode.__table__.c.statys_updated_at,
]

# Example
col = bts.ExecutionNode.__table__.c.statys_updated_at
table_name = col.table.name          # "execution_node"
column_name = col.name                # "statys_updated_at"
column_type = col.type                # DateTime()
```

cloud_pipelines_backend/database_ops.py
```diff
@@ -57,7 +56,30 @@ def create_db_engine(
...
+    inspector = sqlalchemy.inspect(db_engine)
+    for table_name, column_name, column_def in _COLUMN_MIGRATIONS:
+        existing_columns = {col["name"] for col in inspector.get_columns(table_name)}
+        if column_name not in existing_columns:
+            with db_engine.connect() as conn:
+                conn.execute(
+                    sqlalchemy.text(
+                        f"ALTER TABLE {table_name} ADD COLUMN {column_name} {column_def}"
+                    )
+                )
```
yuechao-qin: Let's use DDL for this and try not to use text if possible. For type safety and portability, which is why we're using SQLAlchemy.

```
from sqlalchemy import Column, DateTime
from sqlalchemy.schema import AddColumn
col_obj = bts.ExecutionNode.__table__.c.statys_updated_at
with db_engine.connect() as conn:
    conn.execute(AddColumn("execution_node", col_obj.copy()))
    conn.commit()
```

cloud_pipelines_backend/database_ops.py
```diff
@@ -57,7 +56,30 @@ def create_db_engine(
     return db_engine
 
 
+def _add_columns_if_missing(db_engine: sqlalchemy.Engine) -> None:
```
yuechao-qin: From my backfill lessons learned, i would suggestion

- add logs to know if the migration is happening or not, and any errors
- a single transaction, even if you only have 1 column, for future proofing.  Else remove the for loop for now to make the code align.

Merging all my suggestions into something like this:

```
    _COLUMN_MIGRATIONS = [
        bts.ExecutionNode.__table__.c.statys_updated_at,
    ]
    inspector = sqlalchemy.inspect(db_engine)
    with db_engine.connect() as conn:
        for col in _COLUMN_MIGRATIONS:
            existing = {c["name"] for c in inspector.get_columns(col.table.name)}
            if col.name not in existing:
                _logger.info(
                    f"Migrating: ALTER TABLE {col.table.name} ADD COLUMN {col.name} ({col.type})"
                )
                try:
                    conn.execute(sqlalchemy.schema.AddColumn(col.table.name, col.copy()))
                except sqlalchemy.exc.OperationalError:
                    _logger.info(
                        f"Column {col.table.name}.{col.name} already exists (concurrent migration)"
                    )
            else:
                _logger.info(
                    f"Column {col.table.name}.{col.name} already exists — skipping"
                )
        conn.commit()
```

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -1047,6 +1111,37 @@ def _retry(
     raise
 
 
+def _transition_execution_status(
```
yuechao-qin: Add a * to the function def like mentioned in the other comment.

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -1047,6 +1111,37 @@ def _retry(
...
+) -> None:
+    """Set execution node status and record the status transition duration metric.
+
+    The caller is always responsible for supplying prev_status and
+    prev_status_updated_at.  Typically these are read directly from the live
+    instance immediately before the call.  At call sites that cross a
+    session.rollback() boundary, the values should be captured before the
+    rollback while the instance is still live, as rollback expires all loaded
+    instances and reading an expired attribute would trigger an unintended
+    lazy-load SELECT.
```
yuechao-qin: Curious, for my learning purposes, what does "rollback will expire ..." mean?  Can you give me some examples.  I'm not following the comment.

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -1047,6 +1111,37 @@ def _retry(
...
+    """
+    if prev_status is not None and prev_status_updated_at is not None:
+        app_metrics.record_status_transition(
+            from_status=prev_status.value,
+            to_status=new_status.value,
+            duration_seconds=(
+                _get_current_time() - prev_status_updated_at
+            ).total_seconds(),
+        )
+    execution.container_execution_status = new_status
```
yuechao-qin: this is really cool!

Curious, this should handle ALL node status transitions right?  Asking because I might want to piggyback off this function in the future for doing status search in the search API.

yuechao-qin (changes requested):

cloud_pipelines_backend/instrumentation/metrics.py
```diff
@@ -34,3 +34,24 @@
     description="Number of execution nodes that ended in SYSTEM_ERROR status",
     unit="{error}",
 )
+
+execution_status_transition_duration = orchestrator_meter.create_histogram(
+    name="execution.status_transition.duration",
+    description="Duration an execution spent in a status before transitioning to the next status",
+    unit="s",
```
yuechao-qin: Can this be a strenum? so we know all the units that are possible.

yuechao-qin (changes requested):

cloud_pipelines_backend/backend_types_sql.py
morgan-wowk: Done — renamed to _old_value: object and _initiator: object, and added value: ContainerExecutionStatus | None.

cloud_pipelines_backend/database_ops.py
morgan-wowk: Done — added * to _add_columns_if_missing.

cloud_pipelines_backend/database_ops.py
morgan-wowk: Done — _COLUMN_MIGRATIONS now holds bts.ExecutionNode.**table**.c.status_updated_at and derives table name, column name, and type from it.

cloud_pipelines_backend/database_ops.py
morgan-wowk: `sqlalchemy.schema.AddColumn` isn't available in our SQLAlchemy version, so I kept sqlalchemy.text but compile the type string from the model via col.type.compile(dialect=db_engine.dialect) to stay dialect-portable

cloud_pipelines_backend/database_ops.py
morgan-wowk: Done — adopted your suggestion almost verbatim: single with db_engine.connect() wrapping the loop, an info log before each ALTER TABLE, and a catch for OperationalError on concurrent migration

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Yes!

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Expanded the docstring with a concrete explanation: rollback() marks every attribute on every loaded instance as "expired", so reading prev_status or prev_status_updated_at after a rollback would silently fire a fresh SELECT instead of returning the in-memory value.

cloud_pipelines_backend/instrumentation/metrics.py
morgan-wowk: Done — added MetricUnit(str, enum.Enum) with SECONDS = "s" and ERRORS = "{error}", used for both instruments.

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Done — added * to _transition_execution_status and updated all 16 call sites

cloud_pipelines_backend/database_ops.py
```diff
@@ -66,12 +65,56 @@ def create_db_engine(
...
+    with db_engine.connect() as conn:
+        for col in _COLUMN_MIGRATIONS:
+            existing = {c["name"] for c in inspector.get_columns(col.table.name)}
+            if col.name not in existing:
+                _logger.info(
+                    "Migrating: ALTER TABLE %s ADD COLUMN %s (%s)",
+                    col.table.name,
+                    col.name,
+                    col.type,
+                )
```
yuechao-qin: Let's conform to f-string.  Do the same everywhere (below this function).

Example:

`f"Migrating: ALTER TABLE {col.table.name} ..."`

yuechao-qin (approved):

Ark-kun (changes requested):
Hi, Morgan. Thank you for your patience.
Let's make several changes to improve this feature.

1) Let's follow the Kubernetes example and record history and times of all status transitions.
In some cases Kubernetes records "Last time resource was in state X" to compress the state history when states can be repeated hundreds of times. In our case, we don't expect so much status re-entrance, so let's just record the whole history.
This will allow much richer and future-proof metrics.

Example:
```
container_execution_status_history:
  - status: PENDING
     first_observed_at: XXX
  - status: RUNNING
     first_observed_at: YYY
```

2) Instead of creating a new DB column and adding a manually written migration, let's use `extra_data`. It's designed for the extra data that does not fit into the current static DB schema. This brings SQL tables closer to the usability of the Document DBs.

3) Instead of changing the orchestrator, let's set up a SqlAlchemy event listener for the `ExecutionNode.container_execution_status` change. This allows reliably intercepting all status changes. And it also keeps the orchestrator code simpler.

cloud_pipelines_backend/database_ops.py
morgan-wowk: Done — all variable-interpolated strings in `_add_columns_if_missing` use f-strings. The log calls in `migrate_db` below it are plain literals with no interpolation, so they don't require f-strings.

cloud_pipelines_backend/backend_types_sql.py
```diff
@@ -393,7 +391,75 @@ class ExecutionNode(_TableBase):
         repr=False,
     )
 
-
+    def stage_status_history_entry(
```
Ark-kun: Let's use backend_types to only hold the raw types themselves, not methods working with them.  
It's usually not advised to add business logic methods to ORM classes

cloud_pipelines_backend/backend_types_sql.py
```diff
@@ -393,7 +391,75 @@ class ExecutionNode(_TableBase):
...
+                "%Y-%m-%dT%H:%M:%SZ"
+            ),
+        }
+        pending: dict[str, list] = getattr(self, "_pending_extra_data", None) or {}
+        pending[EXECUTION_NODE_EXTRA_DATA_STATUS_HISTORY_KEY] = pending.get(
+            EXECUTION_NODE_EXTRA_DATA_STATUS_HISTORY_KEY, []
+        ) + [entry]
+        self._pending_extra_data = pending
+
+    def apply_status_history_entries(self) -> None:
```
Ark-kun: I feel like there some overcomplication in these functions.

The pre-flush/post-flush might be needed for telemetry metrics, but should not be needed for nrmal operations where we can just add stuff to extra_data status history structures.

morgan-wowk: Agreed it's overcomplicated. We need to come up with something cleaner.

morgan-wowk:
@Ark-kun This has been simplified per our discussion.

Verified locally in Prometheus:white_check_mark:

cloud_pipelines_backend/sql_event_listeners.py
```diff
@@ -0,0 +1,76 @@
...
+        history: list = (obj.extra_data or {}).get(
+            backend_types_sql.EXECUTION_NODE_EXTRA_DATA_STATUS_HISTORY_KEY, []
+        )
+        if len(history) >= 2:
+            prev = history[-2]
+            curr = history[-1]
+            prev_time = datetime.datetime.fromisoformat(prev["first_observed_at"])
+            curr_time = datetime.datetime.fromisoformat(curr["first_observed_at"])
+            try:
+                metrics.execution_status_transition_duration.record(
```
morgan-wowk: OTel is built to never throw a Runtime exception, but in case it ever did, I don't want that to result in a rollback of the commit - so I've wrapped this with a try/except.

cloud_pipelines_backend/sql_event_listeners.py
```diff
@@ -0,0 +1,76 @@
...
+from .instrumentation import metrics
+
+_logger = logging.getLogger(__name__)
+
+
+@sql_event.listens_for(backend_types_sql.ExecutionNode.container_execution_status, "set")
+def _handle_container_execution_status_set(
+    execution: backend_types_sql.ExecutionNode,
+    value: typing.Any,
+    _old_value: typing.Any,
```
Ark-kun: Do we need to check whether new value is different from the old value?
(If SqlAlchemy already checks that and does not fire event when the value is the same, then no need to change anything.)

cloud_pipelines_backend/sql_event_listeners.py
```diff
@@ -0,0 +1,76 @@
...
+from . import backend_types_sql
+from .instrumentation import metrics
+
+_logger = logging.getLogger(__name__)
+
+
+@sql_event.listens_for(backend_types_sql.ExecutionNode.container_execution_status, "set")
+def _handle_container_execution_status_set(
+    execution: backend_types_sql.ExecutionNode,
+    value: typing.Any,
```
Ark-kun: Maybe let's add type for value?

cloud_pipelines_backend/sql_event_listeners.py
```diff
@@ -0,0 +1,76 @@
...
+    )
+    entry = {
+        "status": value.value,
+        "first_observed_at": datetime.datetime.now(datetime.timezone.utc).strftime(
+            "%Y-%m-%dT%H:%M:%SZ"
+        ),
+    }
+    execution.extra_data = {
+        **execution.extra_data,
+        backend_types_sql.EXECUTION_NODE_EXTRA_DATA_STATUS_HISTORY_KEY: history + [entry],
```
Ark-kun: Let's move this whole function to the orchestrator module. I think maintaining the status history becomes part of it's job, so it belongs there. I'm kind of on the fence here, but modifications to ExecutionNode is probably the jurisdiction of the orchestrator.
This will also solve the issue of wiring it up automatically without relying on an import.

And `_handle_before_commit` function can go to some instrumentation module.

Ark-kun (approved):
Thank you. Approved.

But let's split the sql_event_listeners module (simple).
And let's check that we do not add history entries for duplicate status.

cloud_pipelines_backend/sql_event_listeners.py
morgan-wowk: Added a check.

cloud_pipelines_backend/sql_event_listeners.py
morgan-wowk: Split up. It's not my ideal situation because now the coupled functionality it spread across different modules, but it will do.

cloud_pipelines_backend/sql_event_listeners.py
morgan-wowk: Sounds good. Added

morgan-wowk:
Re-verified with Prometheus after latest changes :white_check_mark:  
  
* Split sql_event_listeners  
* Handled duplicate status

morgan-wowk:
### Merge activity

* **Apr 27, 8:29 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/174).
* **Apr 27, 8:29 PM UTC**: @morgan-wowk merged this pull request with [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/174).

PR #176, opened 2026-03-18
chore: Switch delete user settings endpoint to read from query params

morgan-wowk:
**Changes:**

* Wrap delete user settings handler in `add_parameter_annotation_metadata` to instruct FastAPI to read `setting_names` parameter from query params

### DIFF
--- a/cloud_pipelines_backend/api_router.py
+++ b/cloud_pipelines_backend/api_router.py
@@ -548,10 +548,14 @@ def _setup_routes_internal(
         )
     )
     router.delete("/api/users/me/settings", tags=["user_settings"], **default_config)(
-        inject_session_dependency(
-            inject_user_name(
-                user_settings_service.delete_settings, parameter_name="user_id"
-            )
+        add_parameter_annotation_metadata(
+            inject_session_dependency(
+                inject_user_name(
+                    user_settings_service.delete_settings, parameter_name="user_id"
+                )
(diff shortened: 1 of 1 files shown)


Ark-kun (approved):

morgan-wowk:
The URL would be `http://localhost:8000/api/users/me/settings?setting_names=theme&setting_names=default_compute` to delete multiple keys at the same time.

> Does FastAPI docs support it (schema, UI)

Can you clarify? Do you mean OpenAPI spec and swagger / docusaurus UI? As in, https://oasis.shopify.io/docs

morgan-wowk:
I read through the FastAPI docs and they don't have an explicit preference. There is mention of support for request bodies on DELETE request but no explicit instruction. Gemini, without encouragement, suggests to proceed with using the query parameters.

Ark-kun:
\>Can you clarify? Do you mean OpenAPI spec and swagger / docusaurus UI? As in, https://oasis.shopify.io/docs  
Yes. I wanted to know whether https://oasis.shopify.io/docs and openapi.json shows it ~ the same way as when it's part of the body.  
But this is not a blocker.

morgan-wowk:
I know Swagger supports query parameter definition in general but I'll follow up to make sure this is being translated well from FastAPI. I'll merge for now to not block use of this API.

PR #182, opened 2026-03-20
fix: Add separate GCS key Env for local dev

camielvs:
Overriding the default GCS env variable `GOOGLE_APPLICATION_CREDENTIALS` with GCS signing key results in the app failing to start up. This change adds a secondary env variable to use a GCS signing key `GCS_SIGNING_KEY_FILE`, which is used primary for accessing GCS buckets from local dev.


Ark-kun (approved):

camielvs:
We have opted to pursue other solutions. Closing this PR.

PR #183, opened 2026-03-20
chore: Add CLAUDE.md to register code standards with Claude agent

morgan-wowk:
### TL;DR

Added coding conventions documentation for Python, SQLAlchemy, and data safety practices.

### What changed?

Created `CLAUDE.md` file containing development guidelines including:
- Python style requirements for f-string logging, keyword-only arguments, unused parameter naming, and StrEnum usage
- SQLAlchemy best practices for schema metadata derivation, transaction management, and migration logging
- Data safety rules prohibiting deletion of data files without confirmation

### How to test?

No testing required - this is documentation only. Developers should reference these conventions when writing new code or reviewing pull requests.

### Why make this change?

Establishes consistent coding standards across the repository to improve code quality, maintainability, and prevent data loss incidents during development.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,20 @@
+# Claude Code conventions for this repo
+
+## Python style
+
+- **f-strings for all logging.** Use `_logger.info(f"...")` — never `%s` interpolation.
+- **Keyword-only arguments.** Add `*` to function definitions to make all parameters keyword-only, even when there is only one parameter.
+- **Type hints on all function parameters and return types.** Every function signature must be fully annotated.
+- **StrEnum for known string sets.** When a parameter or constant is drawn from a fixed set of strings, define a `StrEnum` for it rather than using bare string literals.
+- **Imports: modules only.** Import modules, not individual names — except for `from typing import ...`. Example: `import sqlalchemy` not `from sqlalchemy import text`.
+- **File naming: plural or uncountable nouns.** Name Python files as plural or uncountable nouns. Example: `database_migrations.py` not `database_migrate.py`.
+
+## SQLAlchemy
+
+- **Derive schema metadata from models.** Never hardcode table names, column names, or column types as strings. Reference them from the SQLAlchemy model (e.g. `Model.__table__.c.column_name`) and compile types with `col.type.compile(dialect=engine.dialect)`.
+- **Single transaction per logical operation.** Wrap all related DDL or DML mutations in one `with engine.connect() as conn` / `conn.commit()` block.
+- **Log migrations and catch errors.** Every migration step (ALTER TABLE, backfill, index creation) should log before it runs, catch exceptions, log them, and handle `OperationalError` for concurrent deployments gracefully.
+
+## Data safety
+
+- **Never delete or overwrite data files without explicit confirmation.** Use `sqlite://` (in-memory) or a `tempfile` for smoke tests and migration verification — never reference or delete files under `data/`.


CLAUDE.md
```diff
@@ -0,0 +1,18 @@
...
+- **f-strings for all logging.** Use `_logger.info(f"...")` — never `%s` interpolation.
+- **Keyword-only arguments.** Add `*` to function definitions to make all parameters keyword-only, even when there is only one parameter.
+- **Unused parameters.** Prefix unused function parameters with `_` (e.g. `_old_value`, `_initiator`) and add type hints to every parameter, used or not.
+- **StrEnum for known string sets.** When a parameter or constant is drawn from a fixed set of strings, define a `StrEnum` for it rather than using bare string literals.
+
+## SQLAlchemy
+
+- **Derive schema metadata from models.** Never hardcode table names, column names, or column types as strings. Reference them from the SQLAlchemy model (e.g. `Model.__table__.c.column_name`) and compile types with `col.type.compile(dialect=engine.dialect)`.
+- **Single transaction per logical operation.** Wrap all related DDL or DML mutations in one `with engine.connect() as conn` / `conn.commit()` block.
+- **Log migrations.** Every migration step (ALTER TABLE, backfill, index creation) should log before it runs and handle `OperationalError` for concurrent deployments gracefully.
```
yuechao-qin: Catch errors and log them.

CLAUDE.md
```diff
@@ -0,0 +1,18 @@
+# Claude Code conventions for this repo
+
+## Python style
+
+- **f-strings for all logging.** Use `_logger.info(f"...")` — never `%s` interpolation.
+- **Keyword-only arguments.** Add `*` to function definitions to make all parameters keyword-only, even when there is only one parameter.
+- **Unused parameters.** Prefix unused function parameters with `_` (e.g. `_old_value`, `_initiator`) and add type hints to every parameter, used or not.
+- **StrEnum for known string sets.** When a parameter or constant is drawn from a fixed set of strings, define a `StrEnum` for it rather than using bare string literals.
```
yuechao-qin, line 8: - Type hinting function parameters
- Import modules only except for exception of `from typing import ...`
- python filenames should:
```
Style: Modules should be names as plural or uncountable nouns.
Example: database_migrations.py instead of database_migrate.py
```

yuechao-qin (changes requested):

CLAUDE.md
morgan-wowk: Great suggestion — updated the "Log migrations" rule to explicitly include catching and logging exceptions. Thanks!

CLAUDE.md
morgan-wowk: All three added — type hints on every function parameter/return type, module-only imports (with the `from typing import` exception), and plural/uncountable noun file naming. Thanks for the detailed suggestions!

yuechao-qin (approved):

CLAUDE.md
```diff
@@ -0,0 +1,21 @@
+# Claude Code conventions for this repo
+
+## Python style
+
+- **f-strings for all logging.** Use `_logger.info(f"...")` — never `%s` interpolation.
+- **Keyword-only arguments.** Add `*` to function definitions to make all parameters keyword-only, even when there is only one parameter.
+- **Unused parameters.** Prefix unused function parameters with `_` (e.g. `_old_value`, `_initiator`) and add type hints to every parameter, used or not.
```
Ark-kun: Hmm. I'm not sure we [should] do that. Parameter names should not be changed (breaking change) and when creating a function, there should not be any unused parameters. The only case where there can be some unused parameters is when implementing and interface. But in that case the parameter names must match the interface.

There are couple of places where I've used underscored parameters to mean private/internal, although this is not ideal.

morgan-wowk: I will remove this line. It stemmed from an interface implementation I'm pretty sure but I don't remember exactly anymore.

morgan-wowk: The line on unused parameters has been removed.

PR #184, opened 2026-03-21
refactor: Orchestrator - Refactored updating execution status into _record_terminal_state function; Launchers - Kubernetes - Added enum for job condition type

yuechao-qin:
## Background

Kubernetes Job conditions use `"Complete"` for successful jobs and `"Failed"` for failed jobs. The `LaunchedKubernetesJob.ended_at` property was checking for `"Succeeded"` instead of `"Complete"`, which meant `ended_at` was always `None` for successful K8s Jobs. This caused the API to return no end time for completed executions.

Separately, when the orchestrator hit an internal error processing a running execution (SYSTEM_ERROR), it set the status but never recorded `ended_at`, making it impossible to tell when the failure occurred.

## Refactor

- Created enums to set the `JobConditionType` and `ConditionStatus`
- All terminal state fields (`status`, `ended_at`, `exit_code`, `started_at`) are set through a single helper to prevent future inconsistencies

## Changes

**`cloud_pipelines_backend/launchers/kubernetes_launchers.py`**

- Added `JobConditionType` and `ConditionStatus` enums with K8s API doc references, replacing magic strings throughout the file
- Fixed `ended_at` to check `("Complete", "Failed")` instead of `("Succeeded", "Failed")`

**`cloud_pipelines_backend/orchestrator_sql.py`**

- Added `_set_terminal_state()` helper that sets all terminal fields on a `ContainerExecution` in one place
- Refactored SUCCEEDED, FAILED, CANCELLED, and SYSTEM_ERROR branches to use the helper
- SYSTEM_ERROR branch now sets `ended_at` (was previously missing)

## Test

```bash
uv run pytest tests/test_kubernetes_launchers.py tests/test_orchestrator_terminal_state.py
```

### DIFF
--- a/cloud_pipelines_backend/launchers/kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/kubernetes_launchers.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import copy
 import datetime
+import enum
 import json
 import logging
 import os
@@ -68,6 +69,34 @@ _MULTI_NODE_ALL_NODE_ADDRESSES_DYNAMIC_DATA_KEY = "system/multi_node/all_node_ad
 
 # Environment variables for multi-node execution.
 _MULTI_NODE_NODE_INDEX_ENV_VAR_NAME = "_TANGLE_MULTI_NODE_NODE_INDEX"
+
+
+class _JobConditionType(str, enum.Enum):
+    """Kubernetes Job condition types.
+
+    A Job is considered finished when it is in a terminal condition,
+    either "Complete" or "Failed".
+
+    Reference: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/job-v1/
+    See: `A job is considered finished when it is in a terminal condition, either "Complete" or "Failed".`
+    """
+
+    COMPLETE = "Complete"
+    FAILED = "Failed"
+    SUSPENDED = "Suspended"
+    FAILURE_TARGET = "FailureTarget"
+
+
+class _ConditionStatus(str, enum.Enum):
+    """Kubernetes condition status values.
+
+    Reference: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/job-v1/
+    See: `Status of the condition, one of True, False, Unknown.`
+    """
+
+    TRUE = "True"
+    FALSE = "False"
+    UNKNOWN = "Unknown"
 
 
 _T = typing.TypeVar("_T")
@@ -1287,11 +1316,13 @@ class LaunchedKubernetesJob(interfaces.LaunchedContainer):
         if not job_status:
             return interfaces.ContainerStatus.PENDING
         has_succeeded_condition = any(
-            condition.type == "Complete" and condition.status == "True"
+            condition.type == _JobConditionType.COMPLETE
+            and condition.status == _ConditionStatus.TRUE
             for condition in job_status.conditions or []
         )
         has_failed_condition = any(
-            condition.type == "Failed" and condition.status == "True"
+            condition.type == _JobConditionType.FAILED
+            and condition.status == _ConditionStatus.TRUE
             for condition in job_status.conditions or []
         )
         if has_failed_condition:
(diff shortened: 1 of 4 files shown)


cloud_pipelines_backend/launchers/kubernetes_launchers.py
```diff
@@ -67,6 +68,34 @@
 _MULTI_NODE_NODE_INDEX_ENV_VAR_NAME = "_TANGLE_MULTI_NODE_NODE_INDEX"
 
 
+class JobConditionType(str, enum.Enum):
```
Ark-kun: Let's make this private.

cloud_pipelines_backend/launchers/kubernetes_launchers.py
```diff
@@ -67,6 +68,34 @@
...
+    See: `A job is considered finished when it is in a terminal condition, either "Complete" or "Failed".`
+    """
+
+    COMPLETE = "Complete"
+    FAILED = "Failed"
+    SUSPENDED = "Suspended"
+    FAILURE_TARGET = "FailureTarget"
+
+
+class ConditionStatus(str, enum.Enum):
```
Ark-kun: Let's make this private.

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -1010,6 +1021,29 @@ def _get_current_time() -> datetime.datetime:
     return datetime.datetime.now(tz=datetime.timezone.utc)
 
 
+def _set_terminal_state(
```
Ark-kun: I'm not sure this function adds any value. It might also be adding confusion as it's not obvious (from the sall site) what exactly it is doing. Let's revert it.

Ark-kun:
Nit: This looks like a fix combined with a refactoring. This increases the diff and makes it harder to see what has changed and what hasn't. The fix changes are hidden among refactoring changes.

tests/test_orchestrator_terminal_state.py
```diff
@@ -0,0 +1,350 @@
...
+        launcher.get_refreshed_launched_container_from_dict.return_value = refreshed_lc
+
+        storage = mock.MagicMock()
+        orch = _make_orchestrator(launcher=launcher)
+        orch._storage_provider = storage
+        session = _make_mock_session()
+
+        with (
+            mock.patch(
+                "cloud_pipelines_backend.orchestrator_sql._get_current_time",
```
Ark-kun: What happens if we make this a real obejct, not a string? This would be more robust if something is renamed.

tests/test_orchestrator_terminal_state.py
```diff
@@ -0,0 +1,350 @@
...
+            ),
+        ):
+            storage.make_uri.return_value.get_reader.return_value.exists.return_value = (
+                True
+            )
+            info = mock.MagicMock()
+            info.is_dir = False
+            info.total_size = 10
+            info.hashes = {"md5": "abc123"}
+            storage.make_uri.return_value.get_reader.return_value.get_info.return_value = (
```
Ark-kun: This looks brittle. It might be easier to creake a mock storage provider (or a mocked GCS storage provider) and give it to the launcher/orchestrator.

tests/test_orchestrator_terminal_state.py
```diff
@@ -0,0 +1,350 @@
...
+from cloud_pipelines_backend import backend_types_sql as bts
+from cloud_pipelines_backend import orchestrator_sql
+from cloud_pipelines_backend.launchers import interfaces as launcher_interfaces
+
+
+def _utc(*, hour: int = 12) -> datetime.datetime:
+    return datetime.datetime(2026, 3, 20, hour, 0, tzinfo=datetime.timezone.utc)
+
+
+def _make_execution_node(
```
Ark-kun: `_make_mock_execution_node`?

tests/test_orchestrator_terminal_state.py
```diff
@@ -0,0 +1,350 @@
...
+    lc.status = status
+    lc.exit_code = exit_code
+    lc.started_at = started_at
+    lc.ended_at = ended_at
+    lc.launcher_error_message = launcher_error_message
+    lc.to_dict.return_value = {"kubernetes_job": {"refreshed": True}}
+    return lc
+
+
+def _make_mock_session() -> mock.MagicMock:
```
Ark-kun: It might be easier to work with a real in-memory Sqlite DB session than mock it. That would also allow testing more end-to-end with less mocking.

cloud_pipelines_backend/orchestrator_sql.py
yuechao-qin: I messed up on this function. I meant to make the `ended_at` also required, not optional.  So that it's clear for any terminal state transitions it requires 1) status and 2) ended time changes.  I renamed the function to be `_record_terminal_state` and `ended_at` to be required for more clarity.

Do you think this function is helpful then?

Else, do you have any suggestions on how to 1) make it clear to the developers where the terminal transitions are and 2) what required changes (status, ended_at) are needed when transitioning to a terminal state?

Ark-kun:
Let's split the Kubernetes launcher refactor (job condition type enum) and the orchestrator refactor (`_record_terminal_state`).

I support the change in the Kubernetes launcher (after all, I had an error there), but I'm not sure about _record_terminal_state refactor.

PR #186, opened 2026-03-24
feat: Metrics to count all non-terminal execution statuses

morgan-wowk:
### TL;DR

Added a metrics polling service that periodically queries the database to emit execution status count gauges for monitoring active (non-terminal) pipeline executions.

### What changed?

- Added `EXECUTIONS` metric unit to `MetricUnit` enum
- Created `execution_status_count` observable gauge to track execution node counts by status
- Implemented `PollingService` class in new `metrics_poller.py` module that:
    - Queries the database every 30 seconds for execution status counts
    - Only tracks active (non-terminal) statuses like PENDING, RUNNING, etc.
    - Thread-safely updates gauge observations with current counts
- Integrated metrics poller as a daemon thread in the application startup, with automatic detection of OpenTelemetry metrics configuration

### How to test?

1. Set the OpenTelemetry metrics exporter endpoint environment variable (see `examples/observability/otel-jaeger-prometheus`)
2. Start the application and verify the metrics poller thread launches
3. Create pipeline executions in various active statuses
4. Check that the `execution.status.count` gauge emits observations with correct counts for each status
5. Verify that terminal statuses (SUCCEEDED, FAILED) are not included in the gauge metrics

### Why make this change?

This enables real-time monitoring of pipeline execution health by providing visibility into how many executions are currently in each active state, which is essential for operational observability.

### DIFF
--- a/cloud_pipelines_backend/instrumentation/metrics_polling.py
+++ b/cloud_pipelines_backend/instrumentation/metrics_polling.py
@@ -1,0 +1,107 @@
+"""Metrics polling.
+
+Periodically queries the DB and updates ObservableGauges. Currently emits
+execution status counts; add new DB-backed metrics here as needed.
+
+Only fluctuating (non-terminal) statuses are emitted as status count gauges —
+terminal statuses like SUCCEEDED and FAILED only ever climb and are not useful
+as gauges.
+"""
+
+import logging
+import time
+import typing
+
+import sqlalchemy as sql
+from opentelemetry import metrics as otel_metrics
+from sqlalchemy import orm
+
+from .. import backend_types_sql as bts
+from . import metrics as app_metrics
+from .opentelemetry._internal import configuration as otel_configuration
+
+_logger = logging.getLogger(__name__)
+
+
+# All statuses minus terminal (ended) ones — these fluctuate up and down
+_ACTIVE_STATUSES: frozenset[bts.ContainerExecutionStatus] = (
+    frozenset(bts.ContainerExecutionStatus) - bts.CONTAINER_STATUSES_ENDED
+)
+
+
+def _empty_status_counts() -> dict[str, int]:
+    return {s.value: 0 for s in _ACTIVE_STATUSES}
+
+
+class PollingService:
+    """Polls the DB periodically and emits execution status count gauges."""
+
+    def __init__(
+        self,
+        *,
+        session_factory: typing.Callable[[], orm.Session],
+        poll_interval_seconds: float = 30.0,
+    ) -> None:
+        self._session_factory = session_factory
+        self._poll_interval_seconds = poll_interval_seconds
+        # Initialize all active statuses to 0
+        self._counts: dict[str, int] = _empty_status_counts()
+        # Register our observe method as the gauge callback.
+        # The OTel SDK stores callbacks in _callbacks; we append after creation
+        # since create_observable_gauge is called at module load time in metrics.py.
+        app_metrics.execution_status_count._callbacks.append(self._observe)
+
+    def run_loop(self) -> None:
+        while True:
+            try:
+                self._poll()
+            except Exception:
+                _logger.exception("Metrics PollingService: error polling DB")
+            time.sleep(self._poll_interval_seconds)
+
+    def _poll(self) -> None:
+        with self._session_factory() as session:
+            rows = session.execute(
+                sql.select(
+                    bts.ExecutionNode.container_execution_status,
+                    sql.func.count().label("count"),
+                )
+                .where(
+                    bts.ExecutionNode.container_execution_status.in_(_ACTIVE_STATUSES)
+                )
+                .group_by(bts.ExecutionNode.container_execution_status)
+            ).all()
+        new_counts = _empty_status_counts()
+        for status, count in rows:
+            if status is not None:
+                new_counts[status.value] = count
+        # CPython: attribute assignment is atomic under the GIL; no lock needed.
+        # If GIL-free Python is ever adopted, revisit this.
+        self._counts = new_counts
+        _logger.debug(f"Metrics PollingService: polled status counts: {new_counts}")
+
+    def _observe(
+        self, _options: otel_metrics.CallbackOptions
+    ) -> typing.Iterable[otel_metrics.Observation]:
+        counts = self._counts.copy()
+        for status_value, count in counts.items():
+            yield otel_metrics.Observation(count, {"execution.status": status_value})
+
+
+def run(*, db_engine: sql.Engine) -> None:
+    """Check OTel config and run the metrics polling loop (blocking).
+
+    Logs and returns immediately if no metrics endpoint is configured.
+    """
+    otel_config = otel_configuration.resolve()
+    if otel_config is None or otel_config.metrics is None:
+        _logger.info(
+            f"No OTel metrics endpoint configured"
+            f" (set {otel_configuration.EnvVar.METRIC_EXPORTER_ENDPOINT})"
+            f" — metrics poller not starting"
+        )
+        return
+    session_factory = orm.sessionmaker(
+        autocommit=False, autoflush=False, bind=db_engine
+    )
+    PollingService(session_factory=session_factory).run_loop()
--- a/start_local.py
+++ b/start_local.py
@@ -211,6 +211,13 @@ run_configured_orchestrator = lambda: run_orchestrator(
 # endregion
 
 
+# region: Metrics poller initialization
+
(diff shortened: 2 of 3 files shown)


morgan-wowk:
* **#186** <a href="https://app.graphite.com/github/pr/TangleML/tangle/186?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/186?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#174** <a href="https://app.graphite.com/github/pr/TangleML/tangle/174?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

cloud_pipelines_backend/instrumentation/metrics_poller.py
```diff
@@ -0,0 +1,85 @@
...
+        self,
+        *,
+        session_factory: typing.Callable[[], orm.Session],
+        poll_interval_seconds: float = 30.0,
+    ) -> None:
+        self._session_factory = session_factory
+        self._poll_interval_seconds = poll_interval_seconds
+        self._lock = threading.Lock()
+        # Initialize all active statuses to 0
+        self._counts: dict[str, int] = {s.value: 0 for s in _ACTIVE_STATUSES}
```
yuechao-qin: Nit: used in two places, might make sense to put into a function.

cloud_pipelines_backend/instrumentation/metrics_polling.py
```diff
@@ -0,0 +1,85 @@
...
+    ) -> None:
+        self._session_factory = session_factory
+        self._poll_interval_seconds = poll_interval_seconds
+        self._lock = threading.Lock()
+        # Initialize all active statuses to 0
+        self._counts: dict[str, int] = {s.value: 0 for s in _ACTIVE_STATUSES}
+        # Register our observe method as the gauge callback.
+        # The OTel SDK stores callbacks in _callbacks; we append after creation
+        # since create_observable_gauge is called at module load time in metrics.py.
+        app_metrics.execution_status_count._callbacks.append(self._observe)
```
yuechao-qin, line 52: You're calling a private variable, is it possible to make a function set sets up `execution_status_count` with the callbacks in `metric.py`?  

Reason for this is:
- The private variable could change since this is not meant to be called, which will have issues if it does in the future.
- Will there ever be a case you're appending more than one callback?  If so, can you elaborate that use case.  My thought is you would only ever need to "set" one callback.

start_local.py
```diff
@@ -211,6 +211,35 @@ def run_orchestrator(
...
+        return
+    session_factory = orm.sessionmaker(
+        autocommit=False, autoflush=False, bind=db_engine
+    )
+    metrics_poller.PollingService(session_factory=session_factory).run_loop()
+
+
+run_configured_metrics_poller = lambda: run_metrics_poller(
+    db_engine=db_engine,
+)
```
yuechao-qin: It's better to do

```
def run_configured_metrics_poller() -> None:
    run_metrics_poller(db_engine=db_engine)
```

Instead of lambda for variables.  Reason why is 1) troubleshooting, if there was a stack trace it'll just say "lambda" instead of the actual function name and 2) cannot have type hinting/docstrings to have a strongly type function for type checking.

Also, if we had a linter, there are some that would complain about this.

Use cases for when using lambda are inline/throwaway like:
```
sorted(items, key=lambda x: x.created_at)
```

cloud_pipelines_backend/instrumentation/metrics_poller.py
```diff
@@ -0,0 +1,85 @@
...
+            self._counts = new_counts
+        _logger.debug(f"Metrics PollingService: polled status counts: {new_counts}")
+
+    def _observe(
+        self, _options: otel_metrics.CallbackOptions
+    ) -> typing.Iterable[otel_metrics.Observation]:
+        with self._lock:
+            counts = self._counts.copy()
+        for status_value, count in counts.items():
+            yield otel_metrics.Observation(count, {"execution.status": status_value})
```
yuechao-qin: Curious, when will observe be called?  Will there ever be a race condition where observe is called before the poll happened/completed?  Would it make sense to put a flag here that at last 1 poll completed?

start_local.py
morgan-wowk: For now this is intentional to match existing code standards set by the orchestrator. It matches the orchestrator code exactly. I'm not sure I want to introduce an alternative format right now unless @Ark-kun would like make the switch.   
  
I do like all your points for not using a lambda in this case.

cloud_pipelines_backend/instrumentation/metrics_poller.py
morgan-wowk: The lock will prevent it from copying counts that are currently being modified.  
  
As for there being at least 1 iteration, i'll add a check.

cloud_pipelines_backend/instrumentation/metrics_polling.py
morgan-wowk: I'll see what we can do!

start_local.py
```diff
@@ -211,6 +211,35 @@ def run_orchestrator(
...
+            f" (set {otel_configuration.EnvVar.METRIC_EXPORTER_ENDPOINT})"
+            f" — metrics poller not starting"
+        )
+        return
+    session_factory = orm.sessionmaker(
+        autocommit=False, autoflush=False, bind=db_engine
+    )
+    metrics_poller.PollingService(session_factory=session_factory).run_loop()
+
+
```
Ark-kun: Let's move this to metrics_poller

cloud_pipelines_backend/instrumentation/metrics_polling.py
Ark-kun, line 1: Style: Module names should be plural or uncountable nouns. `metrics_polling`?

cloud_pipelines_backend/instrumentation/metrics_poller.py
```diff
@@ -0,0 +1,85 @@
...
+            if status is not None:
+                new_counts[status.value] = count
+        with self._lock:
+            self._counts = new_counts
+        _logger.debug(f"Metrics PollingService: polled status counts: {new_counts}")
+
+    def _observe(
+        self, _options: otel_metrics.CallbackOptions
+    ) -> typing.Iterable[otel_metrics.Observation]:
+        with self._lock:
```
Ark-kun: Are you sure you need lock at all? The dictionary does not seem to ever be modified - it's only ever assigned to new value.

Ark-kun (approved):

cloud_pipelines_backend/instrumentation/metrics_poller.py
morgan-wowk: Good call — extracted into `_empty_status_counts()` module-level helper, used in both `__init__` and `_poll`.

start_local.py
morgan-wowk: Done — moved into `metrics_polling.run(*, db_engine)`. `start_local.py` now just passes `db_engine` and `metrics_polling` handles OTel config detection and session factory creation.

cloud_pipelines_backend/instrumentation/metrics_polling.py
morgan-wowk: Renamed to `metrics_polling.py`. Updated all import sites including oasis-backend.

cloud_pipelines_backend/instrumentation/metrics_poller.py
morgan-wowk: You're right — removed. `self._counts = new_counts` is a single `STORE_ATTR` bytecode — a complete new dict is assigned atomically, so `_observe` never sees a partially-built state. Added a comment noting the CPython GIL assumption.

yuechao-qin (approved):

Ark-kun (approved):

morgan-wowk:
Re-verified after restack :white_check_mark:

morgan-wowk:
### Merge activity

* **Apr 27, 8:29 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/186).
* **Apr 27, 8:29 PM UTC**: [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/186) rebased this pull request as part of a merge.
* **Apr 27, 8:30 PM UTC**: @morgan-wowk merged this pull request with [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/186).

PR #199, opened 2026-04-03
refactor: Move code to appropriate folders for CODEOWNERS

yuechao-qin:
### Overview

Closes https://github.com/Shopify/oasis-frontend/issues/525

Reorganized `cloud_pipelines_backend/` by extracting search, annotation, execution, and backfill logic from the monolithic `api_server_sql.py` into dedicated subpackages (i.e. folders). Tests were split and relocated to mirror the new source structure. No behavioral changes.

### New Directories

| Directory | Purpose |
|-----------|---------|
| `cloud_pipelines_backend/search/` | Filter query models, SQL generation, and pipeline run list/pagination logic |
| `cloud_pipelines_backend/annotation/` | System annotation mirroring, validation guards, and truncation helpers |
| `cloud_pipelines_backend/execution/` | `ExecutionStatusSummary` dataclass |
| `cloud_pipelines_backend/backfill/` | Annotation backfill migrations (renamed from `database_migrations.py`) |

### Why make this change?

The new dedicated folders can be applied to the appropriate `CODEOWNERS`.


cloud_pipelines_backend/execution/utils.py
Ark-kun, line 1: Let's leave it in api_server. Too small.

cloud_pipelines_backend/backfill/annotations.py
Ark-kun, line 1: Let's keep it where it was.

cloud_pipelines_backend/annotation/utils.py
Ark-kun, line 1: Let's move this to search/pipeline_run_annotations.py

Ark-kun, line 1: General style:
All module names should be either plural or uncountable nouns (to avoid clashes with local variables). "search" seems kind of fine. Although if we were a search system dealing with many searches, it would not hav been fine.
Generic catch-all module names like "utils" should be avoided as they tend to accumulate unrelated functionality.

cloud_pipelines_backend/search/runs.py
```diff
@@ -0,0 +1,179 @@
...
+            created_by=pipeline_run.created_by,
+            created_at=pipeline_run.created_at,
+        )
+
+
+@dataclasses.dataclass(kw_only=True)
+class ListPipelineJobsResponse:
+    pipeline_runs: list[PipelineRunResponse]
+    next_page_token: str | None = None
+
```
Ark-kun, line 41: I think that such "response" classes belong to api_server_sql.
I suggest returning pipeline_runs + next page token

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -207,100 +147,16 @@ def list(
...
-        pipeline_run: bts.PipelineRun,
-        include_pipeline_names: bool,
-        include_execution_stats: bool,
-    ) -> PipelineRunResponse:
-        response = PipelineRunResponse.from_db(pipeline_run)
-        if include_pipeline_names:
-            pipeline_name = None
-            extra_data = pipeline_run.extra_data or {}
-            if self._PIPELINE_NAME_EXTRA_DATA_KEY in extra_data:
-                pipeline_name = extra_data[self._PIPELINE_NAME_EXTRA_DATA_KEY]
```
Ark-kun, line 258: Let's keep this code in this module.
The basic functionality unrelated to search should remain here.

Ark-kun (changes requested):

PR #202, opened 2026-04-08
refactor: Move ContainerExecutionStatus to enums module

morgan-wowk:
Breaks the circular import that prevented event_listeners.py from typing
StatusTransitionEvent fields as ContainerExecutionStatus. All consumers
updated to import from the new enums module.

### DIFF
--- a/cloud_pipelines_backend/api_router.py
+++ b/cloud_pipelines_backend/api_router.py
@@ -12,6 +12,7 @@ import starlette.types
 
 from . import api_server_sql
 from . import backend_types_sql
+from . import container_statuses
 from . import component_library_api_server as components_api
 from . import database_ops
 from . import errors
@@ -595,7 +596,7 @@ def _setup_routes_internal(
     )
     def admin_set_execution_node_status(
         id: backend_types_sql.IdType,
-        status: backend_types_sql.ContainerExecutionStatus,
+        status: container_statuses.ContainerExecutionStatus,
         session: typing.Annotated[orm.Session, fastapi.Depends(get_session)],
     ):
         with session.begin():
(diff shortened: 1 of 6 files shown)


morgan-wowk:
* **#203** <a href="https://app.graphite.com/github/pr/TangleML/tangle/203?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#202** <a href="https://app.graphite.com/github/pr/TangleML/tangle/202?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/202?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

cloud_pipelines_backend/api_router.py
```diff
@@ -12,6 +12,7 @@
 
 from . import api_server_sql
 from . import backend_types_sql
+from . import enums
```
yuechao-qin: nit: Would be better to be a bit descriptive for the name here like 
- container_statuses
- models
- statuses

yuechao-qin (approved):

cloud_pipelines_backend/api_router.py
morgan-wowk: Do you suggest a folder instead?  
  
enums  
\\_container_statuses.py

yuechao-qin: not a folder, just renaming the file.  It's just that `enums` and `enum` (python's module) are so close, it's easy for my to confuse the two.

Mainly, it's probably good practice to describe the functionality of the file, not the technology.  For example, if i named a file `classes.py` or `functions.py`, it'll be ambiguous.

morgan-wowk: Thanks. Changed

Ark-kun:
\>Breaks the circular import that prevented event_listeners.py from typing
StatusTransitionEvent fields as ContainerExecutionStatus. 

Hmm. Let's solve this.  
1) There should not be circular imports in this area  
2) Circular imports are possible in Python

PR #203, opened 2026-04-08
feat: Add event_listeners module for event lifecycle support

morgan-wowk:
Provides a general-purpose publish/subscribe system with:
- Event marker base class (enforced via TypeVar bound)
- StatusTransitionEvent dataclass (frozen, kw_only) with typed
  ContainerExecutionStatus fields
- subscribe(event_type, callback) keyed by event class
- emit(event) dispatches to all subscribers by type(event)

### DIFF
--- a/cloud_pipelines_backend/event_listeners.py
+++ b/cloud_pipelines_backend/event_listeners.py
@@ -1,0 +1,45 @@
+import dataclasses
+import threading
+import typing
+
+_CallbackEntry: typing.TypeAlias = tuple[typing.Callable[..., None], bool]
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class Event:
+    """Marker base class for all event types."""
+
+
+_EventType = typing.TypeVar("_EventType", bound=Event)
+
+_listeners: dict[type, list[_CallbackEntry]] = {}
+
+
+def subscribe(
+    *,
+    event_type: type[_EventType],
+    callback: typing.Callable[[_EventType], None],
+    asynchronous: bool = True,
+) -> None:
+    """Subscribe callback to event_type. Called once at startup per consumer.
+
+    Args:
+        event_type: The event class to subscribe to.
+        callback: Called with the event instance when an event of that type is emitted.
+        asynchronous: When True (default), the callback is dispatched on a
+            separate daemon thread so emit() returns immediately. When False,
+            the callback is invoked synchronously on the calling thread.
+    """
+    _listeners.setdefault(event_type, []).append((callback, asynchronous))
+
+
+def emit(
+    *,
+    event: _EventType,
+) -> None:
+    """Dispatch event to all callbacks subscribed to its type."""
+    for callback, asynchronous in _listeners.get(type(event), []):
+        if asynchronous:
+            threading.Thread(target=callback, args=(event,), daemon=True).start()
+        else:
+            callback(event)
--- a/tests/test_event_listeners.py
+++ b/tests/test_event_listeners.py
@@ -1,0 +1,189 @@
+"""Tests for cloud_pipelines_backend.event_listeners."""
+
+import threading
+import typing
+
+import pytest
+
+from cloud_pipelines_backend import event_listeners
+
+
+@pytest.fixture(autouse=True)
+def reset_listeners() -> typing.Generator[None, None, None]:
+    """Clear _listeners before and after every test."""
+    event_listeners._listeners.clear()
+    yield
+    event_listeners._listeners.clear()
+
+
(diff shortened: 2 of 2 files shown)


morgan-wowk:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/TangleML/tangle/203?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#203** <a href="https://app.graphite.com/github/pr/TangleML/tangle/203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#202** <a href="https://app.graphite.com/github/pr/TangleML/tangle/202?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

cloud_pipelines_backend/event_listeners.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class Event:
+    """Marker base class for all event types."""
+
+
+_EventType = typing.TypeVar("_EventType", bound=Event)
+
+_listeners: dict[type, list[_CallbackEntry]] = {}
```
yuechao-qin, line 15: FYI - there is typing.Final which will have linting to make sure this variable can't be assigned.  Not necessary needed here, but just an FYI when I create global variables.

cloud_pipelines_backend/event_listeners.py
```diff
@@ -0,0 +1,45 @@
+import dataclasses
+import threading
+import typing
```
yuechao-qin, line 3: Google style guide for typing would be to not import it as a module, I believe we do that in other places in this repo as well.

https://google.github.io/styleguide/pyguide.html#31912-imports-for-typing

cloud_pipelines_backend/event_listeners.py
```diff
@@ -0,0 +1,45 @@
...
+    """Subscribe callback to event_type. Called once at startup per consumer.
+
+    Args:
+        event_type: The event class to subscribe to.
+        callback: Called with the event instance when an event of that type is emitted.
+        asynchronous: When True (default), the callback is dispatched on a
+            separate daemon thread so emit() returns immediately. When False,
+            the callback is invoked synchronously on the calling thread.
+    """
+    _listeners.setdefault(event_type, []).append((callback, asynchronous))
```
yuechao-qin, line 33: I believe it would be important to thread lock _listeners?  For race conditions between subscribe() and emit()?

If possible, add a new test for this too.

cloud_pipelines_backend/event_listeners.py
```diff
@@ -0,0 +1,45 @@
...
+def emit(
+    *,
+    event: _EventType,
+) -> None:
+    """Dispatch event to all callbacks subscribed to its type."""
+    for callback, asynchronous in _listeners.get(type(event), []):
+        if asynchronous:
+            threading.Thread(target=callback, args=(event,), daemon=True).start()
+        else:
+            callback(event)
```
yuechao-qin, line 45: I believe wrapping a try/catch here would be important to not break the emit for loop?

Add a new test for this too.

yuechao-qin (changes requested):

cloud_pipelines_backend/event_listeners.py
morgan-wowk: I don't think we should worry about such race conditions :thinking_face:. The subscribe action would be synchronous and ideally happen before any user-invoked actions or codepaths are made available that would emit an event to begin with

yuechao-qin: Sounds good!

cloud_pipelines_backend/event_listeners.py
morgan-wowk: Hmm...I guess that would help ensure at least the "async" listeners are guarenteed. But a synchronous listener should prevent any further action / subsequent synchronous listners ffrom executing.

Does that sound good? We add a try catch to ensure the async listeners are always called?

morgan-wowk: We could also just dispatch the asychronous listeners first instead of having a try catch

yuechao-qin (approved):

PR #223, opened 2026-05-04
Add SkyPilot Kubernetes launcher backend for Tangle

Michaelvll:
A new `ContainerTaskLauncher` that submits Tangle tasks as SkyPilot managed
jobs. SkyPilot then handles container scheduling, multi-cluster placement,
multi-node coordination, log mirroring, and cancellation.

## Why

Adding SkyPilot as a launcher backend brings these capabilities to Tangle
pipelines without changing the ComponentSpec authoring model:

- **Multi-cluster / multi-cloud placement.** One launcher instance can dispatch to any context in `kubernetes.allowed_contexts`; sky's optimizer picks per task based on accelerator/region. Lets a single Tangle pipeline fan out across clusters and clouds.
- **Multi-node without a cap**: SkyPilot supports large-scale jobs with no upper limit for the number of nodes
- **High-performance networking**: SkyPilot automatically set up the high-performance infiniband for the jobs without burdens for MLEs
- **Kueue integration (quota + priority).** First-class integration with Kueue -- pipelines can target fair-share / preemption-aware queues directly.
- **Warm-pool reuse.** Submitting to a SkyPilot `Pool` reuses pre-warmed pods instead of creating a fresh one per task, allowing batch inference jobs cutting cold-start from minutes to seconds.
- **Native S3 / R2 / Azure / HTTPS storage mounts.** SkyPilot's `Storage` mounts any of these in addition to GCS — no custom storage-provider code needed in Tangle.
- **Unified logs and cancellation** via `sky.jobs.tail_logs` and `sky.jobs.cancel`, mirrored back into the Tangle UI's `/container_log`.

Opt in:

```bash
pip install tangle[skypilot]

TANGLE_LAUNCHER=skypilot \
  SKYPILOT_API_SERVER_ENDPOINT=http://<sky-api-server> \
  DATABASE_URI=postgresql://... \
  python orchestrator_main.py
```

Existing installs are unaffected; `skypilot` is a new optional dependency group in `pyproject.toml`. Headless / `orchestrator_main.py` users can opt in with `TANGLE_LAUNCHER=skypilot`.

## Examples

23 unit tests pass with sky stubbed (`pytest tests/test_skypilot_launchers.py`).

Live runs targeted Nebius Kubernetes contexts + GCP access with the SkyPilot API server . After starting `start_local_skypilot.py`, the runs below are one `python examples/<name>.py` away.

### 1. Multi-node PyTorch DDP

`examples/multinode_pipeline_e2e.py`. One ComponentSpec annotated with `number_of_nodes: "2"` + `H100:1` runs as a 2-pod NCCL job; rank 0 saves a checkpoint to GCS. SUCCEEDED on 2× NVIDIA H100; loss converged from 3.46 →
0.026 over 5 epochs with identical all-reduced losses across ranks.

**Walkthrough:**

https://github.com/user-attachments/assets/1d6100b8-f175-4eb4-9acb-a130a0611a97

### 2. Inference across multiple clusters

`examples/multicluster_inference_e2e.py`. Four-task graph; one inference
task asks for H100, the other for H200. Sky's optimizer dispatches them to
different K8s contexts in the same pipeline. Same Qwen2.5-0.5B-Instruct
script runs on both, identical greedy completions, a final compare task
fans them together.

**Walkthrough:**

https://github.com/user-attachments/assets/c8c72ec2-c63c-4ccd-9eb2-e73672b2894b

### DIFF
--- a/cloud_pipelines_backend/launchers/skypilot_launchers.py
+++ b/cloud_pipelines_backend/launchers/skypilot_launchers.py
@@ -276,18 +276,28 @@ class SkyPilotKubernetesLauncher(
         job_name = task.name or self._job_name_prefix + "task"
 
         # Submit. sky.jobs.launch returns a RequestId; await it via sky.get().
+        # Result shape changed across sky versions: older returns (List[int], Handle),
+        # newer (>=0.12) returns (Optional[int], Handle). Handle both.
         with self._lock:
             launch_kwargs: dict[str, Any] = {}
             if self._pool is not None:
                 launch_kwargs["pool"] = self._pool
             request_id = sky_jobs.launch(task, name=job_name, **launch_kwargs)
-            job_ids, _handle = sky.get(request_id)
+            result = sky.get(request_id)
 
-        if not job_ids:
+        job_id_or_ids, _handle = result
+        if job_id_or_ids is None:
             raise interfaces.LauncherError(
-                f"sky.jobs.launch returned no job ids for {job_name}"
+                f"sky.jobs.launch returned no job id for {job_name}"
             )
-        job_id = int(job_ids[0])
+        if isinstance(job_id_or_ids, (list, tuple)):
+            if not job_id_or_ids:
+                raise interfaces.LauncherError(
+                    f"sky.jobs.launch returned empty job-id list for {job_name}"
+                )
+            job_id = int(job_id_or_ids[0])
+        else:
+            job_id = int(job_id_or_ids)
         _logger.info(
             "Submitted SkyPilot managed job %s (job_id=%d)", job_name, job_id
         )
@@ -407,6 +417,26 @@ class SkyPilotKubernetesLauncher(
                 for s in ("gs://", "s3://", "abfs://", "https://", "http://", "r2://")
             )
 
+        # SkyPilot's MOUNT mode requires the source to be a bucket root (not a
+        # sub-path within a bucket — sky/data/storage.py:_validate_source raises
+        # StorageModeError otherwise). We mount each unique bucket once at a
+        # stable container path and use sub-paths inside.
+        bucket_mount_root = "/mnt/skypilot"
+
+        def _bucket_and_subpath(uri: str) -> tuple[str, str]:
+            # gs://bucket/sub/path -> ("gs://bucket", "sub/path")
+            scheme, _, rest = uri.partition("://")
+            bucket, _, sub = rest.partition("/")
+            return f"{scheme}://{bucket}", sub
+
+        def _container_path_for(uri: str) -> str:
+            bucket_uri, sub_path = _bucket_and_subpath(uri)
+            scheme = bucket_uri.split("://", 1)[0]
+            bucket_name = bucket_uri.split("://", 1)[1]
+            mount_point = f"{bucket_mount_root}/{scheme}/{bucket_name}"
+            file_mounts[mount_point] = {"source": bucket_uri, "mode": "MOUNT"}
+            return f"{mount_point}/{sub_path}"
+
         def get_input_path(input_name: str) -> str:
             ia = input_arguments[input_name]
             if ia.uri is None:
@@ -415,53 +445,27 @@ class SkyPilotKubernetesLauncher(
                     "storage (gs://, s3://, ...) before submitting through the "
                     "SkyPilot launcher."
                 )
-            container_path = (
-                "/tmp/inputs/"
-                + naming_utils.sanitize_file_name(input_name)
-                + "/"
-                + _CONTAINER_FILE_NAME
-            )
-            if _is_cloud_uri(ia.uri):
-                file_mounts[container_path] = ia.uri
-            else:
-                # Non-cloud URI (e.g. Tangle's LocalStorageProvider relative path):
-                # SkyPilot's file_mounts can't represent this. Surface it clearly
-                # rather than letting sky.Task validation fail with a generic
-                # "file does not exist" error.
+            if not _is_cloud_uri(ia.uri):
                 raise interfaces.LauncherError(
                     f"Input '{input_name}' uri={ia.uri!r} is not a cloud storage URI. "
                     "The SkyPilot launcher requires gs://, s3://, abfs://, https://, or "
                     "r2:// for inputs. Configure Tangle with a cloud StorageProvider "
                     "(e.g. GoogleCloudStorageProvider) for cloud-based runs."
                 )
-            return container_path
+            return _container_path_for(ia.uri)
 
         def get_output_path(output_name: str) -> str:
             uri = output_uris[output_name]
-            container_path = (
-                "/tmp/outputs/"
-                + naming_utils.sanitize_file_name(output_name)
-                + "/"
-                + _CONTAINER_FILE_NAME
-            )
             if _is_cloud_uri(uri):
-                # Bidirectional mount via SkyPilot Storage — the container's writes
-                # land at the cloud URI directly.
-                file_mounts[container_path] = uri
-            else:
-                # Non-cloud (local) output URI. SkyPilot can't sync a remote pod's
-                # writes back to a local relative path. We let the container write to
-                # /tmp/outputs/ (which is ephemeral — the pod terminates after the
-                # job) and log a warning so callers know the artifact won't be
-                # persisted via Tangle's LocalStorageProvider. To persist outputs,
-                # configure a cloud StorageProvider in Tangle.
-                _logger.warning(
-                    "Output '%s' uri=%r is not a cloud URI; the SkyPilot launcher "
-                    "will not persist it back to Tangle's storage. Use a cloud "
-                    "StorageProvider (gs://, s3://, ...) to persist outputs.",
-                    output_name, uri,
-                )
-            return container_path
+                return _container_path_for(uri)
+            _logger.warning(
+                "Output '%s' uri=%r is not a cloud URI; the SkyPilot launcher "
+                "will not persist it back to Tangle's storage. Use a cloud "
+                "StorageProvider (gs://, s3://, ...) to persist outputs.",
+                output_name, uri,
+            )
+            sanitized = naming_utils.sanitize_file_name(output_name)
+            return f"/tmp/outputs/{sanitized}/{_CONTAINER_FILE_NAME}"
 
         resolved = container_component_utils.resolve_container_command_line(
             component_spec=component_spec,
@@ -534,14 +538,22 @@ class SkyPilotKubernetesLauncher(
         elif self._default_use_spot is not None:
             resources_kwargs["use_spot"] = self._default_use_spot
 
-        task = sky.Task(
-            name=job_name,
-            run=run_script,
-            envs=envs,
-            num_nodes=num_nodes,
-            file_mounts=file_mounts or None,
-        )
-        task.set_resources(sky.Resources(**resources_kwargs))
+        # Build a sky YAML-shaped dict and let Task.from_yaml_config parse it.
+        # The YAML parser auto-promotes cloud-URI entries in file_mounts into
+        # sky.Storage MOUNT mounts (sky/task.py:660-688), which is the path that
+        # works under consolidation mode. Constructing sky.Task() directly with
+        # cloud-URI file_mounts goes through translate_local_file_mounts_to_two_hop
+        # which rejects them.
+        task_config: dict[str, Any] = {
+            "name": job_name,
+            "run": run_script,
+            "envs": envs,
+            "num_nodes": num_nodes,
+            "resources": resources_kwargs,
+        }
+        if file_mounts:
+            task_config["file_mounts"] = file_mounts
+        task = sky.Task.from_yaml_config(task_config)
         return task
 
 
@@ -706,8 +718,23 @@ class SkyPilotLaunchedJob(interfaces.LaunchedContainer):
 
     def get_refreshed(self) -> "SkyPilotLaunchedJob":
         request_id = sky_jobs.queue(refresh=True, job_ids=[self._handle.job_id])
-        records = sky.get(request_id)
-        if not records:
+        result = sky.get(request_id)
+        # Sky's queue() return shape varied across versions. Newer (nightly):
+        # tuple[list[dict], ...]; older: list[dict] directly. Unwrap to the
+        # list of records.
+        if isinstance(result, tuple):
+            records = result[0] if result else []
+        else:
+            records = result or []
+        # Find the record matching our job_id (queue may ignore the filter
+        # and return all jobs).
+        rec = None
+        for r in records:
+            jid = r.get("job_id") if isinstance(r, dict) else getattr(r, "job_id", None)
+            if jid == self._handle.job_id:
+                rec = r
+                break
+        if rec is None:
             return SkyPilotLaunchedJob(
                 handle=dataclasses.replace(
                     self._handle,
@@ -715,7 +742,6 @@ class SkyPilotLaunchedJob(interfaces.LaunchedContainer):
                     cached_failure_reason="job not found in sky.jobs.queue",
                 )
             )
-        rec = records[0]
         get = (lambda k: rec.get(k)) if isinstance(rec, dict) else (
             lambda k: getattr(rec, k, None)
         )
--- a/cloud_pipelines_backend/launchers/skypilot_launchers.py
+++ b/cloud_pipelines_backend/launchers/skypilot_launchers.py
@@ -210,11 +210,7 @@ class SkyPilotKubernetesLauncher(
         annotation_to_label_keys: Optional[list[str]] = None,
         priority_class: Optional[str] = None,
         use_spot: Optional[bool] = None,
-        # Default prefix encodes both the orchestrator (Tangle) and the
-        # launcher (SkyPilot). Surfaces in the SkyPilot dashboard, `sky jobs
-        # queue`, and Tangle's debug_info.skypilot.job_name pane — making it
-        # obvious at a glance which launcher ran a given task.
-        job_name_prefix: str = "tangle-skypilot-",
+        job_name_prefix: str = "tangle-",
         storage_provider: Optional[
             storage_provider_interfaces.StorageProvider
         ] = None,
--- a/cloud_pipelines_backend/launchers/skypilot_launchers.py
+++ b/cloud_pipelines_backend/launchers/skypilot_launchers.py
@@ -53,6 +53,7 @@ import json
 import logging
 import shlex
 import threading
+import time
 from typing import Any, Iterator, Optional
 
 from cloud_pipelines.orchestration.launchers import naming_utils
@@ -166,6 +167,9 @@ class _SkyPilotJobHandle:
     cached_ended_at: Optional[float] = None
 
 
+_DEFAULT_DISK_SIZE_GB = 8
+
+
 def _coerce_disk_size_gb(spec: Any) -> int:
     """Best-effort parse of an ephemeral-storage annotation into GiB."""
     if isinstance(spec, (int, float)):
@@ -184,7 +188,11 @@ def _coerce_disk_size_gb(spec: Any) -> int:
     try:
         return max(1, int(float(s)))
     except ValueError:
-        return 8
+        _logger.warning(
+            "Could not parse ephemeral-storage spec %r; defaulting to %d GiB.",
+            spec, _DEFAULT_DISK_SIZE_GB,
+        )
+        return _DEFAULT_DISK_SIZE_GB
 
 
 class SkyPilotKubernetesLauncher(
@@ -192,9 +200,9 @@ class SkyPilotKubernetesLauncher(
 ):
     """Launches Tangle container tasks via SkyPilot managed jobs.
 
-    Designed for Kubernetes-only deployments (the Shopify use case) but works
-    against any infra SkyPilot supports. Set ``infra="kubernetes"`` (or
-    ``"kubernetes/<context>"``) to keep behavior aligned with the existing
+    Defaults to Kubernetes-only dispatch but works against any infra SkyPilot
+    supports. Set ``infra="kubernetes"`` (or ``"kubernetes/<context>"``) to
+    keep behavior aligned with the existing
     KubernetesWithGcsFuseContainerLauncher; pass ``infra=None`` to let
     SkyPilot's optimizer pick across any clouds the user has configured.
     """
@@ -207,7 +215,7 @@ class SkyPilotKubernetesLauncher(
         default_image: Optional[str] = None,
         default_labels: Optional[dict[str, str]] = None,
         default_envs: Optional[dict[str, str]] = None,
-        annotation_to_label_keys: Optional[list[str]] = None,
+        annotation_to_label_keys: Optional[dict[str, str]] = None,
         priority_class: Optional[str] = None,
         use_spot: Optional[bool] = None,
         job_name_prefix: str = "tangle-",
@@ -227,10 +235,13 @@ class SkyPilotKubernetesLauncher(
             default_labels: Labels applied to every Sky resource (propagated to
                 K8s pod labels under the kubernetes infra).
             default_envs: Env vars injected into every container.
-            annotation_to_label_keys: Tangle annotation keys whose values are
-                copied into Sky labels. Useful for passing through things like
-                ``ml.shopify.io/priority-class`` so the K8s pod ends up with the
-                same label kueue is configured to read.
+            annotation_to_label_keys: Map of Tangle annotation key -> Sky label
+                key. For each entry, when the Tangle component annotation is
+                present its value is copied to the named Sky label (and onward
+                to the K8s pod label). Useful for passing through scheduler-
+                relevant hints such as a queue/priority annotation so the K8s
+                pod ends up with the label Kueue (or another admission
+                controller) is configured to read.
             priority_class: Default SkyPilot priority class (Kueue-compatible).
                 Can be overridden per-task via the ``PRIORITY_CLASS_ANNOTATION_KEY``
                 annotation on a ComponentSpec.
@@ -247,7 +258,7 @@ class SkyPilotKubernetesLauncher(
         self._default_image = default_image
         self._default_labels = dict(default_labels or {})
         self._default_envs = dict(default_envs or {})
-        self._annotation_to_label_keys = list(annotation_to_label_keys or [])
+        self._annotation_to_label_keys = dict(annotation_to_label_keys or {})
         self._default_priority_class = priority_class
         self._default_use_spot = use_spot
         self._job_name_prefix = job_name_prefix
@@ -372,11 +383,8 @@ class SkyPilotKubernetesLauncher(
 
         # Labels: defaults, plus selected annotations propagated as labels.
         labels = dict(self._default_labels)
-        for ann_key in self._annotation_to_label_keys:
+        for ann_key, sky_label_key in self._annotation_to_label_keys.items():
             if ann_key in annotations:
-                sky_label_key = (
-                    ann_key.replace("/", "_").replace(".", "_").replace(":", "_")
-                )
                 labels[sky_label_key] = str(annotations[ann_key])
 
         # Pre-resolve multi-node dynamic-data inputs to shell expansions of the
@@ -651,7 +659,6 @@ class SkyPilotLaunchedJob(interfaces.LaunchedContainer):
     _SKY_TERMINAL_STATE_HINT = "is already in terminal state"
 
     def get_log(self) -> str:
-        import time as _time
         # Retry briefly: when the job has just ended, sky's tail_logs returns
         # the terminal-state hint instead of the real log file. The actual
         # logs become available after the controller finalizes them.
@@ -668,7 +675,7 @@ class SkyPilotLaunchedJob(interfaces.LaunchedContainer):
                 # enough content that the hint is just the trailing footer.
                 return out
             if i + 1 < attempts:
-                _time.sleep(delay)
+                time.sleep(delay)
                 delay = min(delay * 1.5, 8.0)
         return out
 
--- a/cloud_pipelines_backend/launchers/skypilot_launchers.py
+++ b/cloud_pipelines_backend/launchers/skypilot_launchers.py
@@ -210,7 +210,11 @@ class SkyPilotKubernetesLauncher(
         annotation_to_label_keys: Optional[list[str]] = None,
         priority_class: Optional[str] = None,
         use_spot: Optional[bool] = None,
-        job_name_prefix: str = "tangle-",
+        # Default prefix encodes both the orchestrator (Tangle) and the
+        # launcher (SkyPilot). Surfaces in the SkyPilot dashboard, `sky jobs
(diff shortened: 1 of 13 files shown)


Ark-kun:
Thank you a lot, Zhanghao! I'll review this PR soon.

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+        return self._handle.cached_failure_reason
+
+    # Sky returns only the "Job N is already in terminal state ..." message
+    # when tail_logs is called immediately after a job ends — the user-job
+    # log file may not yet be flushed/visible on the controller. Detect this
+    # so callers can retry.
+    _SKY_TERMINAL_STATE_HINT = "is already in terminal state"
+
+    def get_log(self) -> str:
+        import time as _time
```
yuechao-qin: Curious, why a lazy import of time?

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+    for suffix, mult in sorted(multipliers.items(), key=lambda kv: -len(kv[0])):
+        if s.endswith(suffix):
+            try:
+                return max(1, int(float(s[: -len(suffix)]) * mult))
+            except ValueError:
+                continue
+    try:
+        return max(1, int(float(s)))
+    except ValueError:
+        return 8
```
yuechao-qin: Should there be a warning log to let user's know user's know it defaulted to 8 GiB?

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+
+
+class SkyPilotKubernetesLauncher(
+    interfaces.ContainerTaskLauncher["SkyPilotLaunchedJob"]
+):
+    """Launches Tangle container tasks via SkyPilot managed jobs.
+
+    Designed for Kubernetes-only deployments (the Shopify use case) but works
```
yuechao-qin: Should "Shopify" be/(need to be) mentioned?

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+
+        # Submit. sky.jobs.launch returns a RequestId; await it via sky.get().
+        # Result shape changed across sky versions: older returns (List[int], Handle),
+        # newer (>=0.12) returns (Optional[int], Handle). Handle both.
+        with self._lock:
+            launch_kwargs: dict[str, Any] = {}
+            if self._pool is not None:
+                launch_kwargs["pool"] = self._pool
+            request_id = sky_jobs.launch(task, name=job_name, **launch_kwargs)
+            result = sky.get(request_id)
```
yuechao-qin, line 308: K8 launchers have timeouts per request ([here](https://github.com/TangleML/tangle/blob/fbb327d59952f6c7be29bd266f86f3124cefc13b/cloud_pipelines_backend/launchers/kubernetes_launchers.py#L179)), is this possible with `sky.get` too?

examples/skypilot/skypilot_launcher_dryrun.py
```diff
@@ -0,0 +1,274 @@
+"""End-to-end dry-run of a Tangle pipeline through the SkyPilot launcher.
+
+Exercises the full ContainerTaskLauncher contract — submit, refresh, log
+streaming, persistence round-trip, terminate — without requiring a live
+Kubernetes cluster or SkyPilot controller. The sky.jobs SDK is stubbed inline
+so the script runs anywhere Tangle is installed.
+
+Run:
+    /home/sky/.venv/bin/python examples/run_pipeline_dryrun.py
+
```
yuechao-qin, line 10: Not sure if I'm doing something wrong here.

Dry-run example broken — `_FakeTask` missing `from_yaml_config`.

`_build_task()` calls `sky.Task.from_yaml_config()` (`skypilot_launchers.py:572`), but the stub in `examples/skypilot_launcher_dryrun.py:35` doesn't implement it. The test stub (`tests/test_skypilot_launchers.py:40`) has it.

Repro:
```
uv sync --group dev --group skypilot
uv pip install -e .
uv run python examples/skypilot_launcher_dryrun.py
```

start_local_skypilot.py
```diff
@@ -0,0 +1,256 @@
...
+def health_check():
+    return {}
+
+
+# Mount the prebuilt frontend (cloned from TangleML/tangle-ui to ./ui_build).
+this_dir = pathlib.Path(__file__).parent
+web_app_search_dirs = [
+    this_dir / "ui_build",
+    this_dir / ".." / "ui_build",
+]
```
yuechao-qin, line 252: Only checks `./ui_build` and `../ui_build`. Missing `../tangle-ui/dist` which `start_local.py` (line 291) includes.

Result: `Frontend build files not found; UI will not be available.`

Fix: add `this_dir / ".." / "tangle-ui" / "dist"` to the search list.

start_local_skypilot.py
```diff
@@ -0,0 +1,256 @@
...
+Run:
+    /home/sky/.venv/bin/uvicorn start_local_skypilot:app --host 0.0.0.0 --port 8000
+
+Or:
+    /home/sky/.venv/bin/python -m uvicorn start_local_skypilot:app --host 0.0.0.0 --port 8000
+
+The Tangle frontend (cloned to ./ui_build) is served at http://localhost:8000/.
+Pipelines submitted from the UI will be dispatched through SkyPilot — they need
+a configured infra (e.g. a Kubernetes context) to actually execute, but the UI
+itself is fully browsable without one.
```
yuechao-qin, line 17: All example scripts hardcode `BASE = "http://localhost:9091"` but docstring says port 8000. Running server on 8000 → examples get `Connection refused`.

Affected files:
- `examples/publish_skypilot_components.py:18`
- `examples/multinode_pipeline_e2e.py:18`
- `examples/multicluster_inference_e2e.py:24`

Suggestion: use `os.environ.get("TANGLE_API_URL", "http://localhost:8000")` so port is configurable and default matches the docstring.

start_local_skypilot.py
```diff
@@ -0,0 +1,256 @@
...
+# Choose between LocalStorageProvider (default) and GoogleCloudStorageProvider
+# via TANGLE_STORAGE_BUCKET env var. SkyPilot's file_mounts can mount cloud
+# URIs (gs://, s3://, abfs://) but cannot represent relative local paths, so
+# multi-step pipelines need a cloud StorageProvider.
+storage_bucket = os.environ.get("TANGLE_STORAGE_BUCKET")
+if storage_bucket:
+    from cloud_pipelines.orchestration.storage_providers import google_cloud_storage
+    storage_provider = google_cloud_storage.GoogleCloudStorageProvider()
+    bucket_uri = storage_bucket.rstrip("/")
+    if not bucket_uri.startswith("gs://"):
```
Ark-kun: We should probably require the URI schema so that we can deduce the storage kind.

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+
+    def __init__(
+        self,
+        *,
+        infra: Optional[str] = "kubernetes",
+        pool: Optional[str] = None,
+        default_image: Optional[str] = None,
+        default_labels: Optional[dict[str, str]] = None,
+        default_envs: Optional[dict[str, str]] = None,
+        annotation_to_label_keys: Optional[list[str]] = None,
```
Ark-kun: Probably better to have it as a map from Tangle annotations to labels.
Speaking of which, you might want a way to map to Kubernetes annotations too.

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+            infra: SkyPilot infra string. ``"kubernetes"`` for any K8s context;
+                ``"kubernetes/<context>"`` to pin to one cluster; ``None`` to
+                let the optimizer pick across all configured clouds.
+            pool: Optional SkyPilot Pool name. Submitting to a warm Pool gives
+                much faster cold-start than full provisioning.
+            default_image: Fallback container image when ComponentSpec doesn't
+                specify one.
+            default_labels: Labels applied to every Sky resource (propagated to
+                K8s pod labels under the kubernetes infra).
+            default_envs: Env vars injected into every container.
```
Ark-kun, line 237: I do not have a strong opinion here. But this might create compatibility issues: The components that rely on extra env variables that are not specified in component or container are not portable.

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+                )
+            return ia.value
+
+        def _is_cloud_uri(uri: str) -> bool:
+            return any(
+                uri.startswith(s)
+                for s in ("gs://", "s3://", "abfs://", "https://", "http://", "r2://")
+            )
+
+        # SkyPilot's MOUNT mode requires the source to be a bucket root (not a
```
Ark-kun, line 444: >SkyPilot's MOUNT mode requires the source to be a bucket root

One problem with this is that it allows faulty user component code to modify or delete all data written by other component task executions.

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+(local_docker_launchers.py, kubernetes_launchers.py).
+
+Storage provider compatibility
+==============================
+
+This launcher relies on SkyPilot's ``file_mounts`` for input/output artifact
+transfer, which can mount cloud-storage URIs (``gs://``, ``s3://``, ``abfs://``,
+``r2://``, ``https://``) directly into the container but cannot represent the
+relative-local-path artifact URIs produced by Tangle's
+``LocalStorageProvider``.
```
Ark-kun, line 18: It's OK for a launcher to only support a subset of storage providers/URIs and raise an exception if unsupported URI is encountered. It's understandable that most launchers won't support the LocalStorageProvider.

P.S. Technically, it's possible to implement (via upload/download data passing instead of mounting). Some clouds that lack good native storage capabilities (e.g. vast.ai) might need that.

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+            file_mounts[mount_point] = {"source": bucket_uri, "mode": "MOUNT"}
+            return f"{mount_point}/{sub_path}"
+
+        def get_input_path(input_name: str) -> str:
+            ia = input_arguments[input_name]
+            if ia.uri is None:
+                raise interfaces.LauncherError(
+                    f"Input '{input_name}' has no URI. Stage values to cloud "
+                    "storage (gs://, s3://, ...) before submitting through the "
+                    "SkyPilot launcher."
```
Ark-kun, line 470: Note that this is not really actionable for the user.
Tangle allows a pipeline to pass a constant value or an task output artifact to any input and also allows a component to consume any input as either value or path.

If the pipeline author wants to provide a constant argument for an input, but component's implementation consumes it as a path, the pipeline author does not have much choice.

So, ideally these "cross" cases should be implemented (OK to do in next PRs).

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+        # which rejects them.
+        task_config: dict[str, Any] = {
+            "name": job_name,
+            "run": run_script,
+            "envs": envs,
+            "num_nodes": num_nodes,
+            "resources": resources_kwargs,
+        }
+        if file_mounts:
+            task_config["file_mounts"] = file_mounts
```
Ark-kun, line 579: Does `file_mounts` support mounting directories as well?

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+            )
+        elif self._default_use_spot is not None:
+            resources_kwargs["use_spot"] = self._default_use_spot
+
+        # Build a sky YAML-shaped dict and let Task.from_yaml_config parse it.
+        # The YAML parser auto-promotes cloud-URI entries in file_mounts into
+        # sky.Storage MOUNT mounts (sky/task.py:660-688), which is the path that
+        # works under consolidation mode. Constructing sky.Task() directly with
+        # cloud-URI file_mounts goes through translate_local_file_mounts_to_two_hop
+        # which rejects them.
```
Ark-kun, line 570: Is this a bug?

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+    def status(self) -> interfaces.ContainerStatus:
+        return _STATUS_MAP.get(
+            self._handle.cached_status or "PENDING", interfaces.ContainerStatus.ERROR
+        )
+
+    @property
+    def exit_code(self) -> Optional[int]:
+        if not self.has_ended:
+            return None
+        return 0 if self.has_succeeded else 1
```
Ark-kun, line 621: Are there plans to make the real exit code available in the future?

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+            component_spec=component_spec,
+            container_spec=container_spec,
+            input_arguments=input_arguments,
+            output_uris=output_uris,
+            annotations=annotations,
+        )
+        job_name = task.name or self._job_name_prefix + "task"
+
+        # Submit. sky.jobs.launch returns a RequestId; await it via sky.get().
+        # Result shape changed across sky versions: older returns (List[int], Handle),
```
Ark-kun, line 301: Is the shape controlled by the Skypilot Python package version of the version of Skypilot used in the cluster?

cloud_pipelines_backend/launchers/skypilot_launchers.py
```diff
@@ -0,0 +1,820 @@
...
+
+        def _run() -> None:
+            try:
+                sky_jobs.tail_logs(
+                    job_id=self._handle.job_id, follow=True, output_stream=buf
+                )
+            finally:
+                finished.set()
+
+        thread = threading.Thread(target=_run, daemon=True)
```
Ark-kun, line 730: I'm a bit concerned about starting new threads in response to the stream_log_lines call.

Is it possible to use `sky_jobs.tail_logs(..., preload_content=True)`? Or is there some issue making that infeasible?

examples/skypilot/multicluster_inference_e2e.py
Ark-kun, line 1: Let's group the examples under a `skypilot` subdirectory.

Ark-kun (approved):
Thank you a lot for this very impressive work!

Apart from a couple of small requests (grouping the examples into skypilot dir; taking a second look at the log streaming), this PR is good to go.

Ark-kun:
@Michaelvll 
I have a few questions about the capabilities.
* Does this launcher support all SkyPilot's cloud providers or are there some limitations?
* How does the "cloud x storage" support matrix look in practice? Is it true that pretty much any bucket can be mounted at any cloud? (I guess, this is a general SkyPilot question.)

yuechao-qin (changes requested):
# Summary of Tests

1. `uv run pytest tests/test_skypilot_launchers.py -v` — 23/23 passed
2. `uv run python examples/publish_skypilot_components.py` — published 2 components successfully
3. Verified published components via `GET /api/published_components/` — returned both digests
4. Verified full component spec via `GET /api/components/{digest}` — YAML matches source
5. `uv run python examples/skypilot_launcher_dryrun.py` — **failed** (missing `from_yaml_config` on stub)

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: No good reason — moved to a module-level `import time`.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: Good call — added a warning log when the ephemeral-storage spec can't be parsed and we fall back to 8 GiB, so it isn't silently swallowed.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: Good point — dropped the Shopify mention from the docstring. The launcher is general-purpose; the K8s-default just mirrors the existing K8s launcher's deployment model.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: `sky.get()` itself doesn't take a timeout (the server holds the connection while resolving the request); there's no client-side knob equivalent to the K8s launcher's per-request timeout. If we need a hard ceiling we'd have to wrap it in a `concurrent.futures.ThreadPoolExecutor.submit(...).result(timeout=...)` here. Happy to add that if you'd like a default — left out of this PR to keep the launcher minimal.

examples/skypilot/skypilot_launcher_dryrun.py
Michaelvll: Thanks for catching this — the dry-run stub was out of sync with the test stub. Fixed by adding `from_yaml_config` to the inline `_FakeTask` in `examples/skypilot/skypilot_launcher_dryrun.py` (mirrors `tests/test_skypilot_launchers.py:40`). Verified the dry-run now runs all 7 sections end-to-end.

start_local_skypilot.py
Michaelvll: Added `../tangle-ui/dist` to the search list.

start_local_skypilot.py
Michaelvll: Fixed — examples now read `TANGLE_API_URL` (defaulting to `http://localhost:8000`, matching the docstring port) so the URL is configurable end-to-end.

start_local_skypilot.py
Michaelvll: Agreed — `TANGLE_STORAGE_BUCKET` now requires a URI scheme and the storage provider is selected from it (currently `gs://` is wired up; other schemes raise a clear error instead of silently defaulting to GCS).

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: Converted `annotation_to_label_keys` to a `dict[str, str]` (Tangle annotation key -> Sky label key) so callers control the final label name explicitly.

Re: Kubernetes pod *annotations* — SkyPilot's `labels` only propagate as K8s pod *labels*, not annotations. SkyPilot doesn't currently expose an `annotations` knob on `Resources` / `Task`; adding one would need an upstream change. Filed as a follow-up — happy to do that and add `annotation_to_pod_annotation_keys` here once the upstream support lands.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: Fair concern. The intent for `default_envs` is launcher-/runtime-level env (credentials helper paths, telemetry endpoints, etc.) that the component shouldn't need to know about — not behavior the container relies on for correctness. Component-relevant env still belongs in `ContainerSpec.env`, which is merged on top of `default_envs`, so a component author opting into a value always wins. I'll add this guidance to the docstring.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: Agreed — mounting at the bucket root gives the task RW access to every other task's data in the same bucket. The mitigation we rely on today is operator-side: configure a Tangle-dedicated bucket and rely on `data_root_uri` / `logs_root_uri` prefixes for per-task path namespacing. Proper per-task scoping would need either (a) bucket-prefix-scoped credentials per task or (b) switching outputs to upload-mode (run-then-upload) so the task never has a writable handle to the whole bucket. Tracking as a follow-up.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: Agreed. For the first cut the launcher just raises on `LocalStorageProvider`. The upload-then-download path you mention is the right direction — it'd let us support clouds (and dev environments) where the running container can't see the orchestrator's filesystem at all. Filing as a follow-up.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: Good point — the cross cases (constant-value-as-path, path-as-constant-value) aren't actionable from the pipeline author. The right place to handle them is the orchestrator, which knows the storage provider and can stage a constant into a temp object (value -> path) or read an output back inline (path -> value). I'll do that in a follow-up; this error currently fires only when the orchestrator handed us an unmaterialized scalar.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: Yes — `file_mounts` mounts at directory granularity. The bucket root is mounted at `/mnt/skypilot/<scheme>/<bucket>/` and any sub-path (file or directory) inside the bucket is reachable as `/mnt/skypilot/<scheme>/<bucket>/<sub_path>`. So a directory input/output works through the same code path — `get_input_path` / `get_output_path` just hand back the sub-path.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: More "quirk" than "bug" — sky has two code paths to populate `file_mounts`:

- `sky.Task(file_mounts=...)` runs through `translate_local_file_mounts_to_two_hop`, which only accepts local paths and rejects cloud URIs.
- `Task.from_yaml_config({"file_mounts": ...})` runs through `_set_file_mounts_from_yaml` (sky/task.py:660-688), which auto-promotes cloud URIs into MOUNT-mode `sky.Storage` mounts.

We need the second path because it's the one that survives consolidation mode. I've left a comment in the code pointing at the relevant sky source lines so it's discoverable; also flagged upstream that the asymmetry is confusing.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: Yes — SkyPilot's managed-job state today only surfaces the high-level `ManagedJobStatus` (SUCCEEDED / FAILED / CANCELLED / FAILED_*), not the underlying job's exit code. I've started a thread upstream on propagating the real exit code through the controller; once it lands, this property will return the actual code instead of the 0/1 placeholder.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: The client-side Python SDK version (which produces the tuple shape from the request response). The server cluster's version matters too — they need to be compatible — but the unwrap logic here is keyed off what the local `sky` package returns. Specifically, in 0.11.x `sky.get(launch_request_id)` resolved to `(List[int], Handle)`; from 0.12 onward it became `(Optional[int], Handle)`. We unwrap both so the launcher works against either pinned version.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: Sky's `tail_logs(follow=True)` is a blocking call that writes to an `output_stream` — it doesn't expose the underlying streaming HTTP response handle (`preload_content=False` is an urllib3 option that sky uses internally but doesn't surface). To convert that blocking writer into a generator we need a producer/consumer split, which means a thread.

Mitigations in the current code:

- Daemon thread, scoped to one `stream_log_lines` call.
- `threading.Event` for completion signaling.
- `join(timeout=1.0)` in the `finally` so we don't leak after the generator exits.

If sky exposes a non-blocking streaming handle in the future, this can collapse into a plain iterator. Open to alternatives if you see a way to avoid the thread — I considered `asyncio.to_thread` but it'd still be the same thread, just managed by asyncio.

examples/skypilot/multicluster_inference_e2e.py
Michaelvll: Done — moved `multicluster_inference_e2e.py`, `multinode_pipeline_e2e.py`, `publish_skypilot_components.py`, and `skypilot_launcher_dryrun.py` to `examples/skypilot/`.

Michaelvll:
Re: cloud-provider / storage support matrix — answering inline here since this came up in the standalone comment:

**Cloud providers.** Every cloud SkyPilot supports is reachable through this launcher, since the launcher just builds a `sky.Task` and delegates to `sky.jobs.launch`. The default `infra="kubernetes"` mirrors the existing K8s launcher's deployment model, but `infra=None` plus a configured `kubernetes.allowed_contexts` lets the optimizer pick across all configured clouds/clusters per task — that's what the `multicluster_inference_e2e.py` example exercises.

**Storage.** SkyPilot's `file_mounts` supports `gs://` / `s3://` / `r2://` / `abfs://` / `https://` mounts uniformly across clouds — there's no "this bucket only works on this cloud" constraint at the SkyPilot layer. A K8s pod on Nebius can mount a GCS bucket; an EC2 instance can mount an R2 bucket; etc.

The practical constraints come from *credentials*, not from sky:

- The runtime needs the matching cloud's credentials available in the pod (e.g. a mounted GCP service-account key for GCS, AWS creds for S3). SkyPilot's helm chart can mount these automatically (`gcpCredentials.enabled=true`, `awsCredentials.enabled=true`, etc.) — the multi-node DDP example does exactly this to write checkpoints from a Nebius pod to GCS.
- Some buckets need cross-cloud-friendly auth (e.g. workload-identity-federation for GCS-from-AWS), which is a cloud-side setup, not a launcher one.

Tangle's storage provider on the orchestrator side (`GoogleCloudStorageProvider` etc.) just needs to be able to produce URIs and serve metadata; the actual data movement happens via sky's `file_mounts` inside the worker pod.

Michaelvll:
Thanks for the careful review, both — pushed [b0989a7](https://github.com/TangleML/tangle/pull/223/commits/b0989a7) addressing the actionable comments:

- **Examples**: moved under `examples/skypilot/` and made the API base URL configurable via `TANGLE_API_URL` (default `http://localhost:8000`, matching the docstring).
- **Dry-run example**: added the missing `from_yaml_config` to the inline `sky.Task` stub so it actually runs end-to-end now.
- **`start_local_skypilot.py`**: include `../tangle-ui/dist` in the frontend search list (parity with `start_local.py`), and require `TANGLE_STORAGE_BUCKET` to include a URI scheme so the storage provider is deduced from the URI.
- **`skypilot_launchers.py`**: dropped the lazy `import time`; added a warning log when ephemeral-storage parsing falls back to the 8 GiB default; removed the Shopify mention from the docstring; converted `annotation_to_label_keys` from `list[str]` to `dict[str, str]` (Tangle annotation key → Sky label key).

Inline replies on each thread cover the design questions (sky.get timeout, MOUNT-mode scoping, LocalStorageProvider, value↔path coercion, file_mounts directory support, YAML `from_yaml_config` asymmetry, exit_code propagation, sky-version-dependent return shape, log-streaming thread). Tests: 23/23 in `tests/test_skypilot_launchers.py` still pass, and the dry-run example runs end-to-end.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Ark-kun: Do you think that mount subPaths can be supported in SkyPilot in the future? This way we mount the data, but only specific sub-folder is accessible to the container code.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Ark-kun: >preload_content=False is an urllib3 option that sky uses internally but doesn't surface
I guess I was looking at the wrong place, but I was reading this code:
https://github.com/skypilot-org/skypilot/blob/2183844f933bdb85e657812a062ac4e5e2b8f652/sky/client/sdk.py#L1006

```
def tail_logs(
    cluster_name: str,
    job_id: Optional[int],
    follow: bool,
    tail: int = 0,
    output_stream: Optional['io.TextIOBase'] = None,
    *,  # keyword only separator
    preload_content: bool = True
) -> Union[int, Iterator[Optional[str]]]:
```

This code had `preload_content` parameter and an Iterator return type (which would be helpful in our case).

Ark-kun (approved):

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: It is on our roadmap for SkyPilot. We will support it. Created an issue for this: https://github.com/skypilot-org/skypilot/issues/9660

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: Created a issue on SkyPilot for supporting mounting the sub directory as well : ) https://github.com/skypilot-org/skypilot/issues/9660

Ark-kun:
Thank you a lot for your contribution, @Michaelvll!

The main follow-up work is handling the constant-value-as-path and path-as-constant-value data passing cases and then potentially simplifying the log streaming and obtaining exit code once those features are more supported in SkyPilot SDK.

We'll be glad to continue collaborating with you in the future.

cloud_pipelines_backend/launchers/skypilot_launchers.py
Michaelvll: You're absolutely right — I was looking at the *managed-jobs* `sky.jobs.tail_logs` and missed that the *cluster-level* `sky.tail_logs` (the one you linked) already exposes `preload_content=False` and returns `Iterator[Optional[str]]`. Apologies for the misleading reply.

The two APIs share the same underlying machinery (`rich_utils.decode_rich_status` over a streaming HTTP response), so the jobs side just hasn't been backfilled with the same iterator-mode escape hatch. Sent the upstream fix here: skypilot-org/skypilot#9662 — it mirrors the cluster overloads + adds `preload_content` to `sky.jobs.tail_logs`. Once that lands, this launcher's `stream_log_lines` becomes a plain `for line in sky_jobs.tail_logs(job_id=..., preload_content=False): yield line` — no thread.

Plan for this PR: leave the current thread-based implementation in place to keep the launcher landable independent of the sky release cadence; switch to `preload_content=False` in a follow-up once skypilot-org/skypilot#9662 ships and we bump the sky pin. Sound good?

yuechao-qin:
@Michaelvll 

Thanks for address my comments.  Retested everything and the following fixes looks good!

- `uv run python examples/skypilot/skypilot_launcher_dryrun.py` — **passed**, all 7 lifecycle steps completed
- `uv run uvicorn start_local_skypilot:app --host 0.0.0.0 --port 8000` — **passed**, UI loads now **AND** defaults to 8000 with Environment Variable override

PR #227, opened 2026-05-07
fix: Fix orchestrator query for (PENDING, RUNNING) by ID only

yuechao-qin:
### TL;DR

Optimized the running executions queue query to avoid loading large JSON columns into MySQL's sort buffer.

### What changed?

The query in `internal_process_running_executions_queue` was split into two steps. Instead of selecting the full `ContainerExecution` row (including potentially large JSON columns) with an `ORDER BY` and `LIMIT`, it now first selects only the `id` column to find the target execution, then fetches the full row by primary key using `session.get()`.

### How to test?

Run the orchestrator against a database with `ContainerExecution` rows that have large JSON column values and verify that running executions are still processed correctly without errors or performance regressions.

### Why make this change?

When MySQL performs a `ORDER BY ... LIMIT` query, it loads all selected columns into the sort buffer. Selecting the entire `ContainerExecution` row — including large JSON columns — was causing the sort buffer to be unnecessarily large. By selecting only the `id` first, the sort operation is performed on a minimal dataset, reducing memory pressure and improving query performance.

#### Sorting by ID only (WORKS)

```
SELECT id
FROM container_execution
WHERE status IN ('PENDING', 'RUNNING')
ORDER BY last_processed_at
LIMIT 1;
```

#### Sorting all (FAILS)

```
SELECT *
FROM container_execution
WHERE status IN ('PENDING', 'RUNNING')
ORDER BY last_processed_at
LIMIT 1;
```

### DIFF
--- a/cloud_pipelines_backend/orchestrator_sql.py
+++ b/cloud_pipelines_backend/orchestrator_sql.py
@@ -131,8 +131,9 @@ class OrchestratorService_Sql:
             return False
 
     def internal_process_running_executions_queue(self, session: orm.Session):
-        query = (
-            sql.select(bts.ContainerExecution)
+        # Select only the ID to avoid loading large JSON columns into the sort buffer.
+        id_query = (
+            sql.select(bts.ContainerExecution.id)
             .where(
                 bts.ContainerExecution.status.in_(
                     (
@@ -144,7 +145,10 @@ class OrchestratorService_Sql:
             .order_by(bts.ContainerExecution.last_processed_at)
             .limit(1)
         )
-        running_container_execution = session.scalar(query)
+        execution_id = session.scalar(id_query)
+        running_container_execution = (
+            session.get(bts.ContainerExecution, execution_id) if execution_id else None
+        )
         if running_container_execution:
             self._running_executions_queue_idle = False
             start_timestamp = time.monotonic_ns()


yuechao-qin:
* **#227** <a href="https://app.graphite.com/github/pr/TangleML/tangle/227?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/227?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -144,7 +145,12 @@ def internal_process_running_executions_queue(self, session: orm.Session):
             .order_by(bts.ContainerExecution.last_processed_at)
             .limit(1)
         )
-        running_container_execution = session.scalar(query)
+        execution_id = session.scalar(id_query)
+        running_container_execution = (
+            session.get(bts.ContainerExecution, execution_id)
+            if execution_id
+            else None
```
morgan-wowk: This is Python backwards reading right? No extra query will happen if execution_id is None?

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -144,7 +145,12 @@ def internal_process_running_executions_queue(self, session: orm.Session):
...
+        )
```
morgan-wowk, line 151: Have you ran the black formatter on this? Just curious.

cloud_pipelines_backend/orchestrator_sql.py
yuechao-qin: Yes, if `execution_id` is `None` then `session.get()` will not be called and `running_container_execution` will be `None`.

morgan-wowk (approved):

cloud_pipelines_backend/orchestrator_sql.py
yuechao-qin: Thank you, I did not, but just ran it and there were formatting changes.

camielvs (approved):

Ark-kun (approved):

PR #233, opened 2026-05-08
Pin exclude-newer to 7-day window in pyproject.toml

morgan-wowk:
### TL;DR

Pins dependency resolution to exclude packages newer than 7 days old.

### What changed?

Added a `[tool.uv]` configuration block to `pyproject.toml` with `exclude-newer = "7 days"`, which instructs `uv` to ignore any packages released within the last 7 days during dependency resolution. The `uv.lock` file has been updated to reflect this constraint with a concrete timestamp cutoff.

### How to test?

Run `uv lock` and verify that the resolved packages in `uv.lock` do not include any releases newer than 7 days from the resolution date.

### Why make this change?

Newly published packages can sometimes introduce unexpected breakage before the broader ecosystem has had a chance to vet them. By excluding packages released within the last 7 days, dependency resolution becomes more stable and less likely to pull in freshly published versions that may contain bugs or breaking changes.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -7,6 +7,10 @@ resolution-markers = [
     "python_full_version < '3.13'",
 ]
 
+[options]
+exclude-newer = "2026-05-01T21:43:53.304008Z"
+exclude-newer-span = "P7D"
+
 [[package]]
 name = "alembic"
 version = "1.18.4"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
     groups:
       minor-patch-updates:
         update-types: ["minor", "patch"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,9 @@ huggingface = [
     "huggingface-hub[oauth]>=0.35.3",
 ]
 
+[tool.uv]
+exclude-newer = "7 days"
+
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 addopts = "--import-mode=importlib"


morgan-wowk:
* **#233** <a href="https://app.graphite.com/github/pr/TangleML/tangle/233?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/233?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

camielvs (approved):

Ark-kun (approved):

uv.lock
```diff
@@ -7,6 +7,10 @@ resolution-markers = [
     "python_full_version < '3.13'",
 ]
 
+[options]
+exclude-newer = "2026-05-01T21:43:53.304008Z"
```
Ark-kun, line 11: I'm a bit concerned about this hardcoded date. It looks like it will add churn.
Under which circumstances will it change?
How will it affect Dependabot?

morgan-wowk: It's generated by uv at lock time. The `exclude-newer` in `pyproject.toml` is a relative `"7 days"` span; uv converts it to an absolute timestamp when writing the lock file's `[options]` section, recording exactly when the lock was last computed.

**When will it change?** Any time uv regenerates the lock: `uv lock`, `uv add`, `uv remove`, or a `uv lock --upgrade-package`. The `[options]` timestamp advances to reflect the new 7-day window. 

It will add minimal churn to each package addition / removal / upgrade. We can expect a 1–2 line diff.

**Dependabot:** Per Astral docs, it works but we should update the dependabot cooldown as well to match our uv `exclude-newer`: https://docs.astral.sh/uv/guides/integration/dependabot/

morgan-wowk: Updated dependabot.yml to have matching behaviour

PR #244, opened 2026-05-20
chore: uv - Refresh uv.lock to include skypilot extras and bump exclude-newer window

Volv-G:
The previous lockfile was missing the skypilot optional dependency group
(53 packages) and its exclude-newer timestamp had drifted outside the
rolling P7D window, both of which break uv sync --locked --all-extras
in CI.

Regenerated against public PyPI with UV_CONFIG_FILE=/dev/null to avoid
picking up the user-level extra-index-url pointing at Shopify's internal
mirror.

PR #246, opened 2026-05-21
chore: uv - Updated uv.lock after adding SkyPilot

Ark-kun:
`uv add --group skypilot skypilot>=0.12.1`


Ark-kun:
Merging Volv's version instead: https://github.com/TangleML/tangle/pull/244

PR #249, opened 2026-05-22
fix: 500 error when published component size is too large

morgan-wowk:
### Problem

`POST /api/published_components/` was crashing with an unhandled `pymysql.err.DataError`
when component text exceeded the MySQL `TEXT` column limit (65,535 bytes):

```
(1406, "Data too long for column 'text' at row 1")
```

The existing `MAX_COMPONENT_SIZE = 300_000` check was in characters, not bytes, and set
far above the actual column capacity — so components between ~65 KB and 300 KB passed
validation and hit the database, which rejected them with a 500.

### Solution

Replace the character-count check with a byte-length check against the real column limit,
and raise `ApiValidationError` (→ HTTP 422) instead of letting the DB error propagate.

**Byte limit**

PostgreSQL and SQLite map `Text()` to unlimited storage, so this limit is based on MySQL.

```python
# Baseline: MySQL TEXT column maximum (65,535 bytes).
# If the column type is ever widened (e.g. to MEDIUMTEXT), update this constant.
MAX_COMPONENT_SIZE_BYTES = 65_535
```

**UTF-8 encoding**

```python
# PyMySQL always encodes Python str to UTF-8 on the wire, so UTF-8 byte length
# is exactly what MySQL receives and measures against the column's byte limit.
text_bytes = len(text.encode("utf-8"))
```

### Changes

- `cloud_pipelines_backend/component_library_api_server.py` — replace `MAX_COMPONENT_SIZE` (character count) with `MAX_COMPONENT_SIZE_BYTES = 65_535` (byte count); raise `errors.ApiValidationError` instead of `ValueError`
- `tests/test_component_library_api_server.py` — add boundary tests for exactly 65,535 bytes (accepted) and 65,536 bytes (rejected with 422)

### DIFF
--- a/cloud_pipelines_backend/component_library_api_server.py
+++ b/cloud_pipelines_backend/component_library_api_server.py
@@ -20,14 +20,19 @@ def calculate_digest_for_component_text(text: str) -> str:
     return digest
 
 
-MAX_COMPONENT_SIZE = 300_000
+# Baseline: MySQL TEXT column maximum (65,535 bytes).
+MAX_COMPONENT_SIZE_BYTES = 65_535
 
 
 def load_component_spec_from_text_and_validate(
     text: str,
 ) -> component_structures.ComponentSpec:
-    if len(text) > MAX_COMPONENT_SIZE:
-        raise ValueError(f"Component size {len(text)} > {MAX_COMPONENT_SIZE=}.")
+    text_bytes = len(text.encode("utf-8"))
+    if text_bytes > MAX_COMPONENT_SIZE_BYTES:
+        raise errors.ApiValidationError(
+            f"Component text is too large: {text_bytes} bytes"
+            f" (maximum allowed: {MAX_COMPONENT_SIZE_BYTES} bytes)."
+        )
     component_dict = yaml.safe_load(text)
     return load_component_spec_from_dict_and_validate(component_dict)
 
--- a/tests/test_component_library_api_server.py
+++ b/tests/test_component_library_api_server.py
@@ -1,4 +1,5 @@
 from sqlalchemy import orm
+import pydantic
 import yaml
 import pytest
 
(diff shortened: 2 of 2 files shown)


morgan-wowk:
* **#249** <a href="https://app.graphite.com/github/pr/TangleML/tangle/249?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/249?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

cloud_pipelines_backend/component_library_api_server.py
```diff
@@ -20,14 +20,19 @@ def calculate_digest_for_component_text(text: str) -> str:
     return digest
 
 
-MAX_COMPONENT_SIZE = 300_000
+# Baseline: MySQL TEXT column maximum (65,535 bytes).
+MAX_COMPONENT_SIZE_BYTES = 65_535
```
yuechao-qin, line 24: Thanks @morgan-wowk for actioning this issue!

Tangle is suppose to be DB agnostic, and this limit is for MySQL while (Postgres and SQLite) doesn't have this limit.  I'm wondering if we should:
1. Move this change to oasis-backend, since that would be proper for KateSQL (MySQL) only.
2. Create a dialect aware helper class in Tangle (too complex IMHO, and I don't think we have a precendent for that, nor would it be wise to create a precendent now).
3. Something else?

## TEXT Column Size Limits by Database

| Database | TEXT Max Size | Effectively Unlimited? |
|----------|-------------|----------------------|
| **MySQL** | 65,535 bytes (~64 KB) | No |
| **PostgreSQL** | ~1 GB | Yes |
| **SQLite** | ~1 GB (default) | Yes |

### MySQL

> "A TEXT column with a maximum length of 65,535 (2^16 − 1) bytes. The effective maximum length is less if the value contains multibyte characters."

**Source**: [MySQL 8.4 Reference Manual - String Data Type Syntax](https://dev.mysql.com/doc/refman/8.4/en/string-type-syntax.html)

### PostgreSQL

> "In addition, PostgreSQL provides the `text` type, which stores strings of any length."
>
> "the longest possible character string that can be stored is about 1 GB"

**Source**: [PostgreSQL Docs - Character Types](https://www.postgresql.org/docs/current/datatype-character.html)

### SQLite

> "The maximum length of a TEXT or BLOB in bytes."
> `#define SQLITE_MAX_LENGTH 1000000000`

**Source (source header)**: [sqliteLimit.h](https://sqlite.org/src/doc/tip/src/sqliteLimit.h)

> "The maximum number of bytes in a string or BLOB in SQLite is defined by the preprocessor macro SQLITE_MAX_LENGTH. The default value of this macro is 1 billion (1,000,000,000)."

**Source (docs)**: [SQLite Implementation Limits](https://sqlite.org/limits.html)

yuechao-qin (changes requested):

cloud_pipelines_backend/component_library_api_server.py
morgan-wowk: Hey @yuechao-qin  :wave:  
  
Regarding the couple of options:  
  
\> Move this change to oasis-backend  
  
While it would help Shopify, I'd like to solve this issue for everyone (OSS included).  
  
\> Create a dialect aware helper class in Tangle   
  
This also feels a bit counter to being database-agnostic but definitely an understandable idea for a problem that is database-specific. I also think it's not the right time to go in that direction.  
  
How about an os.getenv variable that defaults to 65_535 but can be tuned by the OSS consumer according to their database requirements?

yuechao-qin: @morgan-wowk I like the ENV variable idea. 

- If there isn't an ENV variable, default should be 65 KB
- Log something helpful about the component size limit.

However, before you do any changes, please review with Alexey to see which path he wants.

morgan-wowk:
**Note** **from** **meeting:** We decided we would like to increase our MySQL column size rather than making the limit more strict.

PR #261, opened 2026-05-28
bugsnag: strip JSON objects from grouping strings, normalize LauncherError

morgan-wowk:
Any embedded JSON/dict literal (starting with `{"` or `{'`) is now
stripped from error messages before they are used as grouping keys.
This prevents Kubernetes pod specs and similar runtime blobs from
fragmenting what is structurally the same error into many groups.

A dedicated normalizer for `LauncherError` is also added: it drops
everything after the first colon (the serialized pod spec) so that
`"Failed to create pod: <spec>"` normalizes to just
`"LauncherError: Failed to create pod"`.

### DIFF
--- a/cloud_pipelines_backend/instrumentation/error_normalization.py
+++ b/cloud_pipelines_backend/instrumentation/error_normalization.py
@@ -9,6 +9,14 @@ regardless of which specific resource was involved.
 import json
 import re
 
+try:
+    from ..launchers.interfaces import LauncherError as _LauncherError
+
+    _LAUNCHER_ERROR_AVAILABLE = True
+except ImportError:
+    _LauncherError = None  # type: ignore[assignment,misc]
+    _LAUNCHER_ERROR_AVAILABLE = False
+
 _POD_NAME_PATTERN = re.compile(r"(?:task|tangle(?:-ce)?)-[a-zA-Z0-9]+-[a-zA-Z0-9]+")
 _OBJECT_REPR_PATTERN = re.compile(r"<[^>]+ object at 0x[0-9a-fA-F]+>")
 _HEX_ADDRESS_PATTERN = re.compile(r"\b0x[0-9a-fA-F]+\b")
@@ -16,9 +24,15 @@ _UUID_PATTERN = re.compile(
     r"[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}", re.IGNORECASE
 )
 _LONG_ALNUM_ID_PATTERN = re.compile(r"\b[a-zA-Z0-9]{16,}\b")
+# Matches from the first `{"`, `{'`, or `{ "` / `{ '` to end of string.
+# Both the embedded dict/JSON literal and any trailing message text are replaced
+# with `{...}` — the greedy match is intentional: anything after a runtime-data
+# dict in an error message is typically also variable and should not affect grouping.
+_JSON_OBJECT_PATTERN = re.compile(r"\{\s*['\"].*", re.DOTALL)
 
 
 def _strip_generic(*, message: str) -> str:
+    message = _JSON_OBJECT_PATTERN.sub("{...}", message)
     message = _OBJECT_REPR_PATTERN.sub("{object}", message)
     message = _HEX_ADDRESS_PATTERN.sub("{addr}", message)
     message = _UUID_PATTERN.sub("{uuid}", message)
@@ -85,6 +99,13 @@ def _normalize_orchestrator_error(*, exception: BaseException) -> str | None:
     return f"OrchestratorError: {message}"
 
 
+def _normalize_launcher_error(*, exception: BaseException) -> str | None:
+    if not _LAUNCHER_ERROR_AVAILABLE or not isinstance(exception, _LauncherError):
+        return None
+    message = _JSON_OBJECT_PATTERN.sub("{...}", str(exception))
+    return f"LauncherError: {message.strip()}"
+
+
 def normalize_error_message(*, exception: BaseException) -> str:
     """Return a stable normalized string for error grouping."""
     for normalizer in (
@@ -92,6 +113,7 @@ def normalize_error_message(*, exception: BaseException) -> str:
         _normalize_max_retry_error,
         _normalize_unicode_decode_error,
         _normalize_orchestrator_error,
+        _normalize_launcher_error,
     ):
         result = normalizer(exception=exception)
         if result is not None:
--- a/tests/instrumentation/test_error_normalization.py
+++ b/tests/instrumentation/test_error_normalization.py
@@ -184,6 +184,52 @@ class TestNormalizeOrchestratorError:
         )
 
 
+class TestNormalizeLauncherError:
+    def _make_launcher_error(
+        self, message: str, cause: BaseException | None = None
+    ) -> Exception:
+        try:
+            from cloud_pipelines_backend.launchers.interfaces import LauncherError
+        except ImportError:
+            pytest.skip("LauncherError not importable")
+        if cause:
+            try:
+                raise LauncherError(message) from cause
+            except LauncherError as exc:
+                return exc
+        return LauncherError(message)
+
+    def test_strips_pod_spec_json(self):
+        pod_spec = (
+            "{'apiVersion': 'v1', 'kind': 'Pod', 'metadata': {'name': 'task-abc-xyz'}}"
+        )
+        exc = self._make_launcher_error(f"Failed to create pod: {pod_spec}")
+        result = error_normalization.normalize_error_message(exception=exc)
+        assert result == "LauncherError: Failed to create pod: {...}"
+
+    def test_with_timeout_cause(self):
+        cause = TimeoutError("The read operation timed out")
+        exc = self._make_launcher_error(
+            "Failed to create pod: {'apiVersion': 'v1'}", cause=cause
+        )
+        result = error_normalization.normalize_error_message(exception=exc)
+        assert result == "LauncherError: Failed to create pod: {...}"
+
+    def test_no_colon_in_message(self):
+        exc = self._make_launcher_error("launch failed")
+        result = error_normalization.normalize_error_message(exception=exc)
+        assert result == "LauncherError: launch failed"
+
+    def test_multi_colon_diagnostic_preserved(self):
+        exc = self._make_launcher_error(
+            "creating pod: spec invalid: missing field 'name'"
+        )
+        result = error_normalization.normalize_error_message(exception=exc)
+        assert (
+            result == "LauncherError: creating pod: spec invalid: missing field 'name'"
+        )
+
+
 class TestFallback:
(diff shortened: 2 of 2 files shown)


morgan-wowk:
* **#268** <a href="https://app.graphite.com/github/pr/TangleML/tangle/268?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#262** <a href="https://app.graphite.com/github/pr/TangleML/tangle/262?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#261** <a href="https://app.graphite.com/github/pr/TangleML/tangle/261?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/261?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

cloud_pipelines_backend/instrumentation/error_normalization.py
```diff
@@ -16,9 +16,14 @@
...
 )
 _LONG_ALNUM_ID_PATTERN = re.compile(r"\b[a-zA-Z0-9]{16,}\b")
+# Matches any embedded JSON object or Python dict literal (starts with `{"` or `{'`).
+# These are stripped from grouping strings because they contain highly variable
+# runtime data (e.g. full Kubernetes pod specs) that would fragment error groups.
+_JSON_OBJECT_PATTERN = re.compile(r"\{['\"].*", re.DOTALL)
 
 
 def _strip_generic(*, message: str) -> str:
+    message = _JSON_OBJECT_PATTERN.sub("{...}", message)
```
Mbeaulne, line 35: This should literally replace the `_JSON_OBJECT_PATTERN` with `{...}` and not  just `...` correct?

cloud_pipelines_backend/instrumentation/error_normalization.py
```diff
@@ -85,13 +90,27 @@ def _normalize_orchestrator_error(*, exception: BaseException) -> str | None:
...
+def _normalize_launcher_error(*, exception: BaseException) -> str | None:
+    try:
+        from ..launchers.interfaces import LauncherError
+    except ImportError:
+        return None
+    if not isinstance(exception, LauncherError):
+        return None
+    # Take only the verb phrase before the first colon to drop any embedded
+    # serialized data (e.g. the full Kubernetes pod spec appended after ": ").
+    head = str(exception).split(":", 1)[0].strip()
```
Mbeaulne: > 🤖 This is an AI-generated code review comment.

**`_normalize_launcher_error` over-truncates and may be redundant.**

```python
head = str(exception).split(":", 1)[0].strip()
return f"LauncherError: {head}"
```

Splitting on the first colon collapses meaningfully distinct failures into one group. All of these become `LauncherError: Failed to create pod`:
- `"Failed to create pod: connection timeout to API server"`
- `"Failed to create pod: name 'task-x' already exists"`
- `"Failed to create pod: insufficient memory in node pool"`
- `"Failed to create pod: {full pod spec}"`

These are different operational problems and should group separately.

With the new `_JSON_OBJECT_PATTERN` already in `_strip_generic`, the generic fallback would produce `LauncherError: Failed to create pod: {...}` — keeping the verb phrase, marking the JSON-bearing suffix, and still grouping the spec-variable case. The specialized normalizer may not be needed at all.

The comment also calls it "the verb phrase," but for `"creating pod: spec invalid: missing field 'name'"` the head is just `"creating pod"` and the actual diagnostic is dropped.

**Fix:** either (a) drop `_normalize_launcher_error` entirely and let the generic path handle it now that `_JSON_OBJECT_PATTERN` exists, or (b) strip only the trailing JSON-y data and keep the rest:

```python
message = _JSON_OBJECT_PATTERN.sub("{...}", str(exception))
return f"LauncherError: {message.strip()}"
```

cloud_pipelines_backend/instrumentation/error_normalization.py
```diff
@@ -16,9 +16,14 @@
     r"[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}", re.IGNORECASE
 )
 _LONG_ALNUM_ID_PATTERN = re.compile(r"\b[a-zA-Z0-9]{16,}\b")
+# Matches any embedded JSON object or Python dict literal (starts with `{"` or `{'`).
+# These are stripped from grouping strings because they contain highly variable
+# runtime data (e.g. full Kubernetes pod specs) that would fragment error groups.
+_JSON_OBJECT_PATTERN = re.compile(r"\{['\"].*", re.DOTALL)
```
Mbeaulne: > 🤖 This is an AI-generated code review comment.

**`_JSON_OBJECT_PATTERN` greedily consumes everything after the first `{\"`/`{'` to end-of-string — worth documenting.**

```python
_JSON_OBJECT_PATTERN = re.compile(r"\{['\"].*", re.DOTALL)
```

With `re.DOTALL` and the greedy `.*`, this matches from the first `{\"`/`{'` to end-of-string. Trailing context after an embedded JSON literal is also dropped, so:
```
RuntimeError("config: {'a': 1} during step X")  →  RuntimeError: config: {...}
RuntimeError("config: {'a': 1} during step Y")  →  RuntimeError: config: {...}
```
These group together even though "step X" vs "step Y" is a meaningful distinguisher.

This is likely intentional (anything after a runtime-data dict in an error message is usually also variable), but the comment explains the why for the JSON itself, not the trailing-text trade-off. One line in the docstring would help the next reader reason about edge cases — something like "Note: trailing message text after the first `{\"`/`{'` is also stripped".

If the trade-off is not desired, scope the match more tightly with a non-greedy variant like `\{['\"].*?\}` (with caveats around nested braces).

tests/instrumentation/test_error_normalization.py
```diff
@@ -184,6 +184,41 @@ def test_strips_object_repr(self):
...
+        exc = self._make_launcher_error("Failed to create pod: {big spec}", cause=cause)
+        result = error_normalization.normalize_error_message(exception=exc)
+        assert result == "LauncherError: Failed to create pod"
+
+    def test_no_colon_in_message(self):
+        exc = self._make_launcher_error("launch failed")
+        result = error_normalization.normalize_error_message(exception=exc)
+        assert result == "LauncherError: launch failed"
+
+
```
Mbeaulne, line 232: > 🤖 This is an AI-generated code review comment.

**Consider adding a test for multi-colon `LauncherError` messages.**

The three new tests cover JSON suffix, cause chain, and no-colon. None covers the case where there is real diagnostic info after the first colon — exactly the scenario where the verb-phrase truncation silently degrades grouping quality (see the other comment on `_normalize_launcher_error`).

Adding:
```python
def test_multiple_colons_in_message(self):
    exc = self._make_launcher_error("creating pod: spec invalid: missing field 'name'")
    result = error_normalization.normalize_error_message(exception=exc)
    assert result == "LauncherError: creating pod"  # or whatever the desired behavior is
```

…either pins the truncation as intentional or surfaces the regression so the policy gets revisited.

cloud_pipelines_backend/instrumentation/error_normalization.py
```diff
@@ -85,13 +90,27 @@ def _normalize_orchestrator_error(*, exception: BaseException) -> str | None:
     return f"OrchestratorError: {message}"
 
 
+def _normalize_launcher_error(*, exception: BaseException) -> str | None:
+    try:
+        from ..launchers.interfaces import LauncherError
```
yuechao-qin: Wondering if it would make sense to do the import globally (at the top of the file) and have a global bool to track if LauncherError is available?

cloud_pipelines_backend/instrumentation/error_normalization.py
morgan-wowk: 🤖 Yes — the pattern replaces the matched span with `{...}`, which includes the opening `{`. So the result is `{...}` not `...`.

Also improved the pattern to handle optional whitespace after `{` (e.g. `{ "key": ...}`) and updated the comment to document that the greedy match intentionally strips everything from the first dict/JSON literal to end-of-string.

cloud_pipelines_backend/instrumentation/error_normalization.py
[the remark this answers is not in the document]
morgan-wowk: 🤖 Good catch on both points. The greedy behaviour is intentional — anything after an embedded dict/JSON literal in an error message is typically also runtime-variable, so stripping to end-of-string is the right call for grouping stability. The comment has been updated to document this explicitly.

The whitespace fix (`\{\s*`) has also been applied so `{ "key": ...}` now matches correctly.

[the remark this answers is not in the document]
morgan-wowk: 🤖 Fixed — pattern is now `\{\s*['"].*` so optional whitespace after `{` is handled.

cloud_pipelines_backend/instrumentation/error_normalization.py
morgan-wowk: 🤖 Agreed — went with option (b). `_normalize_launcher_error` now applies `_JSON_OBJECT_PATTERN.sub` to the full message and keeps all the diagnostic context before the JSON blob. `Failed to create pod: {'spec': ...}` becomes `LauncherError: Failed to create pod: {...}` and `creating pod: spec invalid: missing field 'name'` passes through unchanged. Tests updated to match.

cloud_pipelines_backend/instrumentation/error_normalization.py
morgan-wowk: 🤖 Updated the comment above `_JSON_OBJECT_PATTERN` to explicitly document that trailing text is also stripped (intentional — anything after a runtime dict is typically also variable).

tests/instrumentation/test_error_normalization.py
morgan-wowk: 🤖 Added `test_multi_colon_diagnostic_preserved` which asserts `creating pod: spec invalid: missing field 'name'` passes through unmodified, confirming that diagnostic colons are kept now that we no longer truncate at the first colon.

cloud_pipelines_backend/instrumentation/error_normalization.py
morgan-wowk: 🤖 Good suggestion! Moved to a module-level try/except with a `_LAUNCHER_ERROR_AVAILABLE` bool flag, so the import is attempted once at startup rather than on every `_normalize_launcher_error` call. The function now checks the flag directly.

Mbeaulne (approved):

morgan-wowk:
### Merge activity

* **Jun 13, 12:04 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/261).
* **Jun 13, 12:04 AM UTC**: @morgan-wowk merged this pull request with [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/261).

PR #262, opened 2026-05-28
bugsnag: chain-aware grouping key and chain title in error list

morgan-wowk:
`normalize_error_chain` walks `__cause__`/`__context__` and joins each level's normalized string with `" <- "`. This replaces the single-exception `normalize_error_message` call used for the grouping key so that `"LauncherError <- TimeoutError"` and `"LauncherError <- ApiException"` land in separate, stable groups rather than collapsing into one group per root cause.

`build_chain_title` does the same walk but truncates each level to 80 chars and returns `None` for single exceptions. `_before_notify` uses it to override the root-cause `errorClass` that Bugsnag displays in the error list, giving reviewers the full raise path at a glance without needing to open the stacktrace.

### DIFF
--- a/cloud_pipelines_backend/instrumentation/bugsnag_instrumentation.py
+++ b/cloud_pipelines_backend/instrumentation/bugsnag_instrumentation.py
@@ -43,19 +43,35 @@ def _before_notify(event: bugsnag_event.Event) -> None:
     if context:
         event.add_tab("tangle_context", context)
     if _CUSTOM_GROUPING_KEY and event.original_error:
-        normalized = error_normalization.normalize_error_message(
+        # Use the full chain for grouping so that "LauncherError <- TimeoutError"
+        # and "LauncherError <- ApiException" land in separate, stable groups.
+        chain = error_normalization.normalize_error_chain(
             exception=event.original_error
         )
         prefix = (event.metadata.get("extra") or {}).get("grouping_prefix")
-        key_value = f"{prefix}: {normalized}" if prefix else normalized
+        key_value = f"{prefix}: {chain}" if prefix else chain
         event.add_tab("custom", {_CUSTOM_GROUPING_KEY: key_value})
         if prefix and event.errors:
             try:
                 for error in event.errors:
                     error.error_class = f"{prefix}: {error.error_class}"
             except Exception:
-                _logger.debug(
+                _logger.warning(
                     "Could not prepend grouping prefix to errorClass", exc_info=True
+                )
+        # For chained exceptions, override the root-cause title (what Bugsnag
+        # displays in the error list) with the full chain so reviewers can see
+        # the complete raise path without opening the stacktrace.
+        chain_title = error_normalization.build_chain_title(
+            exception=event.original_error
+        )
+        if chain_title and event.errors:
+            try:
+                title = f"{prefix}: {chain_title}" if prefix else chain_title
+                event.errors[-1].error_class = title
+            except Exception:
+                _logger.warning(
+                    "Could not set chain title on errorClass", exc_info=True
                 )
 
 
--- a/cloud_pipelines_backend/instrumentation/error_normalization.py
+++ b/cloud_pipelines_backend/instrumentation/error_normalization.py
@@ -107,7 +107,7 @@ def _normalize_launcher_error(*, exception: BaseException) -> str | None:
 
 
 def normalize_error_message(*, exception: BaseException) -> str:
-    """Return a stable normalized string for error grouping."""
+    """Return a stable normalized string for a single exception (no chain traversal)."""
     for normalizer in (
         _normalize_k8s_api_exception,
         _normalize_max_retry_error,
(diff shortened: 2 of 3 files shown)


morgan-wowk:
* **#268** <a href="https://app.graphite.com/github/pr/TangleML/tangle/268?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#262** <a href="https://app.graphite.com/github/pr/TangleML/tangle/262?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/262?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#261** <a href="https://app.graphite.com/github/pr/TangleML/tangle/261?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

cloud_pipelines_backend/instrumentation/error_normalization.py
```diff
@@ -117,3 +117,46 @@ def normalize_error_message(*, exception: BaseException) -> str:
...
+def build_chain_title(*, exception: BaseException) -> str | None:
+    """Return a human-readable chain title for display, or ``None`` for single exceptions.
+
+    Like ``normalize_error_chain`` but truncates each level so the result fits
+    in a Bugsnag error list title.  Returns ``None`` when there is no
+    chain so callers can skip overriding the default title.
+    """
+    parts: list[str] = []
+    seen: set[int] = set()
+    exc: BaseException | None = exception
```
Mbeaulne: > 🤖 This is an AI-generated code review comment.

**Duplicated chain-walking loop between `normalize_error_chain` and `build_chain_title`.**

Both functions share identical traversal logic — same `seen` set, same `len(parts) < 4` cap, same `__cause__ or __context__` walk. The only divergence is the per-part truncation step and the `None` return for single exceptions.

Two places that have to stay in lockstep on every future change (max depth, `BaseExceptionGroup` support, additional cycle detection, etc.). Easy to drift.

**Fix:** extract one private helper:
```python
def _walk_chain(exception: BaseException, max_depth: int = 4) -> list[BaseException]:
    excs: list[BaseException] = []
    seen: set[int] = set()
    exc: BaseException | None = exception
    while exc is not None and id(exc) not in seen and len(excs) < max_depth:
        seen.add(id(exc))
        excs.append(exc)
        exc = exc.__cause__ or (None if exc.__suppress_context__ else exc.__context__)
    return excs
```
Then `normalize_error_chain` is one comprehension and `build_chain_title` is one comprehension with the truncation step.

cloud_pipelines_backend/instrumentation/error_normalization.py
```diff
@@ -117,3 +117,46 @@ def normalize_error_message(*, exception: BaseException) -> str:
...
+
+    Walks ``__cause__`` (and ``__context__`` when not suppressed) and joins
+    each level with `` <- ``.  Use this for grouping keys so that chained
+    exceptions like ``LauncherError <- TimeoutError`` produce one stable group
+    rather than one per root cause.
+    """
+    parts: list[str] = []
+    seen: set[int] = set()
+    exc: BaseException | None = exception
+    while exc is not None and id(exc) not in seen and len(parts) < 4:
```
Mbeaulne: > 🤖 This is an AI-generated code review comment.

**Grouping key produced by `normalize_error_chain` is length-unbounded — Bugsnag may silently truncate.**

```python
parts.append(normalize_error_message(exception=exc))  # no per-part cap
```

This function is used as the Bugsnag custom grouping key. Each part has no length bound — only `build_chain_title` caps at `_CHAIN_PART_MAX_LEN`. With 4 chained exceptions each carrying a long `_strip_generic`-processed message, the grouping key can run into the thousands of characters. If Bugsnag silently truncates custom grouping keys beyond some limit, two chains that differ only past the truncation point group together — exactly the opposite of this PR's intent.

**Fix:** either (a) verify Bugsnag's actual custom-grouping-key length limit and document it in a comment, or (b) apply a generous per-part cap inside `normalize_error_chain` too — e.g. 200 chars, separate from the display-oriented 80-char cap. The cap is stable across chain orderings and protects against accidental key blow-up.

cloud_pipelines_backend/instrumentation/bugsnag_instrumentation.py
```diff
@@ -57,6 +59,18 @@ def _before_notify(event: bugsnag_event.Event) -> None:
...
+        # the complete raise path without opening the stacktrace.
+        chain_title = error_normalization.build_chain_title(
+            exception=event.original_error
+        )
+        if chain_title and event.errors:
+            try:
+                title = f"{prefix}: {chain_title}" if prefix else chain_title
+                event.errors[-1].error_class = title
+            except Exception:
+                _logger.debug("Could not set chain title on errorClass", exc_info=True)
```
Mbeaulne: > 🤖 This is an AI-generated code review comment.

**DEBUG-level logging hides SDK API drift in the title override.**

```python
except Exception:
    _logger.debug("Could not set chain title on errorClass", exc_info=True)
```

The code is built on a load-bearing assumption that `event.errors[-1].error_class` is the field Bugsnag displays as the error-list title. If a future SDK changes that shape, the assignment will throw `AttributeError`, the chain title silently disappears from Bugsnag, and the DEBUG log will be invisible in any production logging config. This is the kind of integration drift that doesn't break unit tests but quietly degrades the product.

**Fix:** log at `warning` (or `error`) level. Still non-fatal — the original error still reports — but the next time the SDK shape drifts, you find out instead of guessing why the error list looks the way it used to.

Note: the same DEBUG pattern at the prefix-prepending block above (~lines 55-59) has the same issue; worth bumping that too, although it's pre-existing.

yuechao-qin (approved):

cloud_pipelines_backend/instrumentation/error_normalization.py
morgan-wowk: 🤖 Extracted `_walk_chain` — both `normalize_error_chain` and `build_chain_title` now delegate to it. The traversal logic (max depth, cycle detection, cause/context walk) lives in one place.

cloud_pipelines_backend/instrumentation/error_normalization.py
morgan-wowk: 🤖 Added a `_CHAIN_GROUPING_KEY_MAX_PART_LEN = 200` cap applied inside `normalize_error_chain`. Keeps individual parts from blowing up the grouping key while staying well above the 80-char display cap used by `build_chain_title`.

cloud_pipelines_backend/instrumentation/bugsnag_instrumentation.py
morgan-wowk: 🤖 Bumped both DEBUG calls in `_before_notify` to WARNING — the prefix-prepend block and the chain title block. SDK shape drift will now surface in production logs rather than disappearing silently.

Mbeaulne (approved):

morgan-wowk:
### Merge activity

* **Jun 13, 12:04 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/262).
* **Jun 13, 12:06 AM UTC**: [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/262) rebased this pull request as part of a merge.
* **Jun 13, 12:06 AM UTC**: @morgan-wowk merged this pull request with [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/262).

PR #267, opened 2026-05-28
bugsnag: emit slice_name from service_name in custom metadata tab

morgan-wowk:
Some Bugsnag stacks filter on custom.slice_name to assign errors to
a slice. Store the service_name passed to setup() and write it as
custom.slice_name on every event so errors from orchestrator, api, and
metrics_poller each land in their own slice.

Consolidates all custom tab writes into a single add_tab call.

PR #268, opened 2026-05-28
bugsnag: propagate cloud_provider into contextual logging for queued executions

morgan-wowk:


morgan-wowk:
* **#268** <a href="https://app.graphite.com/github/pr/TangleML/tangle/268?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/268?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#262** <a href="https://app.graphite.com/github/pr/TangleML/tangle/262?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#261** <a href="https://app.graphite.com/github/pr/TangleML/tangle/261?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

morgan-wowk:
CC: @yuechao-qin

This will allow you to filter errors to Nebius or GKE.

cloud_pipelines_backend/instrumentation/contextual_logging.py
```diff
@@ -125,3 +129,21 @@ def logging_context(**metadata: Any):
...
+    """Return a logging context populated with metadata for *execution*.
+
+    Always sets ``execution_id``.  Also sets ``cloud_provider`` when the
+    ``cloud-pipelines.net/orchestration/cloud_provider`` annotation is present
+    on the task spec.
+    """
+    ctx: dict[str, str] = {"execution_id": execution.id}
+    cloud_provider = (
+        (execution.task_spec or {})
+        .get("annotations", {})
```
Mbeaulne: > 🤖 This is an AI-generated code review comment.

**`annotations: None` (key present, value None) makes the chained `.get` crash.**

```python
cloud_provider = (
    (execution.task_spec or {})
    .get("annotations", {})           # returns None, NOT {}, if annotations is None
    .get(_CLOUD_PROVIDER_ANNOTATION_KEY)
)
```

`.get("annotations", {})` only uses the default when the key is *missing*. If `task_spec = {"annotations": None}` (key present, value None — common in JSON-roundtripped data where keys are emitted even when unset), `.get("annotations", {})` returns `None` and the trailing `.get(...)` raises `AttributeError: 'NoneType' object has no attribute 'get'`. The crash propagates out of `execution_logging_context`, breaking the `with` block setup at the call site and silently degrading log correlation for that execution.

**Fix:** chain `or {}` after the inner `.get` too:
```python
cloud_provider = (
    ((execution.task_spec or {}).get("annotations") or {})
    .get(_CLOUD_PROVIDER_ANNOTATION_KEY)
)
```

cloud_pipelines_backend/instrumentation/contextual_logging.py
```diff
@@ -25,6 +25,10 @@
 from contextlib import contextmanager
 from typing import Any, Optional
 
+from .. import backend_types_sql as bts
```
Mbeaulne: > 🤖 This is an AI-generated code review comment.

**New `instrumentation` → core import is a circular-import hazard.**

```python
from .. import backend_types_sql as bts
```

The `instrumentation` package was previously dependency-free w.r.t. core model code — `bugsnag_instrumentation.py`, `error_normalization.py`, and the rest don't import from `..`. This is the first instrumentation → core import. If `backend_types_sql` ever wants to call back into instrumentation for logging (a plausible future change — e.g. logging inside an `__post_init__` or validator), you get an import cycle.

The only runtime use of `bts` here is the type annotation on `execution: bts.ExecutionNode`. That doesn't need a real import.

**Fix:** guard with `TYPE_CHECKING`:
```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from .. import backend_types_sql as bts
```
`bts.ExecutionNode` becomes a forward reference resolved by the type checker, with zero runtime cost and no cycle risk.

cloud_pipelines_backend/instrumentation/contextual_logging.py
```diff
@@ -125,3 +129,21 @@ def logging_context(**metadata: Any):
     finally:
         # Restore previous metadata
         _context_metadata.set(prev_metadata)
+
+
+def execution_logging_context(execution: bts.ExecutionNode):
```
Mbeaulne: > 🤖 This is an AI-generated code review comment.

**No unit tests for `execution_logging_context`.**

The new function has several non-trivial branches: `task_spec is None`, `task_spec` without `annotations`, `annotations` without the cloud-provider key, the key set to a string, the key set to `None`, the key with an empty string. None of these is pinned by a test, and the orchestrator call site relies on the helper being correct.

**Fix:** add a handful of cases in `tests/instrumentation/test_contextual_logging.py`:
```python
def test_execution_logging_context_no_annotations():
    exc = make_execution_node(id="e1", task_spec={})
    with contextual_logging.execution_logging_context(exc):
        assert contextual_logging.get_context_metadata("execution_id") == "e1"
        assert contextual_logging.get_context_metadata("cloud_provider") is None

def test_execution_logging_context_with_cloud_provider():
    exc = make_execution_node(
        id="e1",
        task_spec={"annotations": {"cloud-pipelines.net/orchestration/cloud_provider": "gcp"}},
    )
    with contextual_logging.execution_logging_context(exc):
        assert contextual_logging.get_context_metadata("cloud_provider") == "gcp"

def test_execution_logging_context_annotations_is_none():
    # Regression guard for None handling (see other comment).
    exc = make_execution_node(id="e1", task_spec={"annotations": None})
    with contextual_logging.execution_logging_context(exc):
        assert contextual_logging.get_context_metadata("cloud_provider") is None
```

cloud_pipelines_backend/instrumentation/contextual_logging.py
```diff
@@ -25,6 +25,10 @@
 from contextlib import contextmanager
 from typing import Any, Optional
 
+from .. import backend_types_sql as bts
+
+_CLOUD_PROVIDER_ANNOTATION_KEY = "cloud-pipelines.net/orchestration/cloud_provider"
```
Mbeaulne, line 32: > 🤖 This is an AI-generated code review comment.

**Module docstring's metadata-key list should mention `cloud_provider`.**

After this PR, `cloud_provider` starts appearing in logs for queued executions. The module docstring at the top of this file enumerates the common metadata keys (`request_id`, `pipeline_run_id`, `execution_id`, etc.) — but `cloud_provider` isn't listed. People grepping the docstring to learn "what metadata keys exist?" won't find it.

**Fix:** add one line to the docstring's "Common metadata keys" list:
```
- cloud_provider: From task_spec annotations, set by execution_logging_context for orchestrated runs.
```

yuechao-qin, line 32: nebius_launchers.py has this same exact constant.  Would it make sense to create a shared config/model/etc file that both of them reference against?  Or some way to not duplicate the variable?

yuechao-qin (dismissed):

cloud_pipelines_backend/instrumentation/contextual_logging.py
morgan-wowk: 🤖 Fixed — the chain is now `((execution.task_spec or {}).get("annotations") or {}).get(...)` so a `None`-valued `annotations` key is treated the same as an absent key. Added a regression test `test_no_cloud_provider_when_annotations_is_none` to cover this case.

cloud_pipelines_backend/instrumentation/contextual_logging.py
morgan-wowk: 🤖 Guarded with `TYPE_CHECKING` — the import is now `if TYPE_CHECKING: from .. import backend_types_sql as bts` and the function signature uses a string annotation `"bts.ExecutionNode"`. Zero runtime cost, no cycle risk.

cloud_pipelines_backend/instrumentation/contextual_logging.py
morgan-wowk: 🤖 Added `tests/instrumentation/test_contextual_logging.py` with six cases: always-sets-execution-id, annotation absent, annotation present with value, task_spec None, annotations None (regression for the crash fix), and context cleared after the block exits.

cloud_pipelines_backend/instrumentation/contextual_logging.py
morgan-wowk: 🤖 Added `cloud_provider` to the module docstring's "Common metadata keys" list with a brief description of where it comes from.

morgan-wowk: 🤖 Good eye, but the duplication is intentional for now. The tech lead asked us not to commit `CLOUD_PROVIDER_ANNOTATION_KEY` to the shared `common_annotations` layer yet — there's an open question about whether a launcher-specific annotation is the right long-term standard, or whether a different annotation design would be better. Once there's consensus, we can consolidate. Until then, keeping it private to each module avoids prematurely locking in the API.

Mbeaulne (approved):

Ark-kun (approved):

morgan-wowk:
Smoke-tested E2E locally.

morgan-wowk:
### Merge activity

* **Jun 13, 12:04 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/268).
* **Jun 13, 12:09 AM UTC**: [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/268) rebased this pull request as part of a merge.
* **Jun 13, 12:10 AM UTC**: @morgan-wowk merged this pull request with [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/268).

PR #271, opened 2026-06-04
fix: GCS composite object has no MD5 hash

yuechao-qin:
## Summary

The orchestrator crashes with `AttributeError: 'NoneType' object has no attribute 'encode'` when collecting output artifacts from GCS that are [composite objects](https://cloud.google.com/storage/docs/composite-objects). Composite objects (created by parallel uploads or compose operations) have no MD5 hash — `blob.md5_hash` is `None`.

The crash happens in the upstream SDK (`cloud-pipelines==0.26.3.12`) at `GoogleCloudStorageProvider._get_info_from_uri`, during output collection after a container completes. This is before any data reaches our code, so there's no way to fix it at the save/consume layer.

## Fix

- **New file**: `patched_google_cloud_storage.py` — subclasses the SDK's `GoogleCloudStorageProvider` and overrides `_get_info_from_uri` to handle `None` MD5 gracefully
- When MD5 is missing, substitutes `"no_md5_<UTC timestamp>"` as the hash value — this forces a cache miss for any downstream task consuming that artifact (Solution A: safe, no false cache hits)
- **Updated launchers** (`google_kubernetes_launchers.py`, `kubernetes_launchers.py`) to use `PatchedGoogleCloudStorageProvider` instead of the upstream provider

## Failing runs

- https://oasis.shopify.io/runs/019e8e17de7041a0dd8c
- https://oasis.shopify.io/runs/019e8f7be22321296b8a

## Trade-offs

- Composite object outputs will **always cause a cache miss** for downstream tasks — this is intentional and safe
- Long-term fix: compute our own MD5 by downloading the file (cost scales with file size)

## Tests

7 unit tests added covering:
- Normal blobs return correct MD5 hex digest
- Composite blobs return unique timestamp fallback
- Timestamps are unique across calls
- Warning is logged for composite objects
- Provider handles single files, composite files, and mixed directories

### DIFF
--- a/cloud_pipelines_backend/launchers/google_kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/google_kubernetes_launchers.py
@@ -2,7 +2,7 @@ import typing
 
 from kubernetes import client as k8s_client_lib
 
-from cloud_pipelines.orchestration.storage_providers import google_cloud_storage
+from cloud_pipelines_backend.storage_providers import patched_google_cloud_storage
 
 from . import kubernetes_launchers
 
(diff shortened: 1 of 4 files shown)


yuechao-qin:
* **#271** <a href="https://app.graphite.com/github/pr/TangleML/tangle/271?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/271?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #277, opened 2026-06-17
fix: Launchers - Kubernetes - Fix getting logs when logs arte not valid UTF-8

### LINKED ISSUE #281: UnicodeDecodeError in pod log reading marks healthy GPU training runs as SYSTEM_ERROR
## Root cause

When a Kubernetes pod completes a GPU training run with tqdm progress bars, the pod log contains multi-byte UTF-8 block glyphs (`█▉▊▋▌▍▎▏`, each 3 bytes: `E2 96 8x`). Two distinct mechanisms produce torn byte sequences in the raw log bytes that the Kubernetes API returns:

1. **Concurrent writer interleaving** — 64 worker processes share a single stderr fd. Python's `TextIOWrapper` can split one logical `write()` call across multiple `write()` syscalls. When two processes' writes interleave at byte granularity, the receiver sees an orphaned continuation byte (`invalid continuation byte`).

2. **Response truncation at a chunk boundary** — the K8s API response is split at a size boundary (~5.1 MB observed), cutting a 3-byte glyph after its first byte (`unexpected end of data` at `0xe2`).

The Kubernetes Python client reads the response with `r.data.decode('utf8')` (strict, no error handling). Both variants throw `UnicodeDecodeError` before the run has any real failure. The exception bubbles through the `_retry` wrapper in `internal_process_one_running_execution` and, after 5 retries, marks the execution `SYSTEM_ERROR` — causing downstream Upload steps to be skipped on an otherwise successful training run.

**This is responsible for 3,960 Bugsnag occurrences in the past 7 days.**

Neither the Linux kernel nor the Kubernetes client is at fault. The kernel's `PIPE_BUF` atomic write guarantee applies only to writes ≤ 4096 bytes, and Python's `TextIOWrapper` may issue multiple syscalls for one logical write. The corruption is baked into the bytes before the K8s client sees them, so the client cannot fix it in a lossless way.

## Solutions compared

Three approaches were kept as separate PRs for team comparison:

**PR #277 — Defensive decode (`_preload_content=False` + `errors="replace"`)**
Bypass the K8s client's strict decode by using `_preload_content=False`, then decode ourselves with `errors="replace"` so `U+FFFD` is substituted for any malformed byte sequence. A `_logger.warning` fires when substitution occurs.
- Pro: handles both error classes; logs remain complete (only the torn glyph is replaced)
- Con: lossy — the torn bytes (partial tqdm glyph) are replaced rather than recovered

**PR #278 — Suppress tqdm output at source**
Inject `TQDM_DISABLE=1` (and related env vars) into the pod's environment so block glyphs are never written to stderr in the first place.
- Pro: eliminates the root cause entirely for tqdm-based progress bars
- Con: only covers tqdm; any other multi-byte output from training code or libraries would still tear

**PR #279 / #280 — Swallow log-acquisition errors + Observe search link**
Catch all exceptions from log reading, return an empty string / `None`, and embed a pre-built Observe log search URL in the placeholder message so the user can still find the pod logs.
- Pro: execution is never incorrectly marked `SYSTEM_ERROR`; user has a direct link to raw logs
- Con: the execution log stored in GCS is lost; relies on Observe retention

## Fix landed

**PR #277** — `_preload_content=False` + `errors="replace"` was selected. It addresses both UnicodeDecodeError variants, keeps the execution log intact (minus the torn glyph bytes), and does not require changes to the training environment.

Fixes #277

morgan-wowk:
The Kubernetes client's default read_namespaced_pod_log path does a strict
.decode('utf8') over the full log payload before checking HTTP status.
When a pod with high-volume tqdm progress bars (block glyphs █▉▊▋▌▍▎▏,
3-byte UTF-8) runs with num_proc>1, concurrent writes to the same fd can
split a multi-byte glyph across a chunk boundary, leaving an orphaned
continuation byte.  The strict decode throws UnicodeDecodeError, which
bubbles through the log-upload retry wrapper and marks an otherwise-healthy
training run as SYSTEM_ERROR.

Fix: pass _preload_content=False to get the raw urllib3 response and decode
manually with errors="replace".  This is applied to both the single-pod
(LaunchedKubernetesContainer.get_log) and multi-pod
(LaunchedKubernetesJob._get_log_by_pod_key) log-read paths.
A warning is logged whenever replacement characters are injected, so the
next occurrence is observable in Observe without requiring a separate
debug build.

The existing "Bad Request" catch for PodInitializing is unaffected:
the kubernetes client's status check runs outside the _preload_content
block and still raises ApiException with the correct reason phrase.

### Orchestrator log-upload path (run lifecycle)

**Before** — the UnicodeDecodeError bubbles out of the retry wrapper. The run is marked `SYSTEM_ERROR`, no logs are uploaded, and all downstream tasks (e.g. Upload HF, Upload Training Summary) are skipped. The user sees a failed run with no log output and no indication that their training code was healthy.

**After** — the log is decoded successfully and uploaded. The run continues to completion. One or two progress-bar characters are substituted with `?` (U+FFFD) at the point of corruption, but the rest of the log is intact and readable.

### API log-read path (viewing logs for a running execution)

**Before** — the request throws before returning a response. The user gets a 500 error in the UI when trying to view logs mid-run.

**After** — the full log is returned. The substituted character appears inline exactly where the torn byte was, typically mid-progress-bar where it is visually unnoticeable.

---

### Example log output

**Before** (UnicodeDecodeError thrown at byte 5,115,152 — nothing returned):
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe2 in position 5115152: unexpected end of data
```

**After** (log returned; `?` marks the single substituted byte at the corruption point):
```
2024-06-17T20:29:11Z Map (num_proc=64):  77%|██████████████████████████████████████████████████████████████████████████████████████▌         | 137412/178432 [00:34<00:10, 3991.03 examples/s]
2024-06-17T20:29:12Z Map (num_proc=64):  79%|███████████████████████████████████████████████████████████████?██████  | 140876/178432 [00:35<00:09, 3987.11 examples/s]
2024-06-17T20:29:13Z Map (num_proc=64):  81%|█████████████████████████████████████████████████████████████████████▏  | 144501/178432 [00:36<00:08, 3994.77 examples/s]
2024-06-17T20:29:55Z ***** Running training *****
2024-06-17T20:29:55Z   Num examples = 144,501
2024-06-17T20:29:55Z   Num Epochs = 3
```

The `?` on line 2 is where one torn block glyph was replaced. All structured log lines above and below it — training config, loss values, eval metrics — are fully intact.

Fixes: https://github.com/TangleML/tangle/issues/281

### DIFF
--- a/cloud_pipelines_backend/launchers/kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/kubernetes_launchers.py
@@ -922,7 +922,8 @@ class LaunchedKubernetesContainer(interfaces.LaunchedContainer):
     def get_log(self) -> str:
         launcher = self._get_launcher()
         core_api_client = k8s_client_lib.CoreV1Api(api_client=launcher._api_client)
-        return core_api_client.read_namespaced_pod_log(
+        # _preload_content=False bypasses the kubernetes client's strict .decode('utf8'); see _get_log_by_pod_key.
+        response = core_api_client.read_namespaced_pod_log(
             name=self._pod_name,
             namespace=self._namespace,
             container=_MAIN_CONTAINER_NAME,
@@ -931,7 +932,17 @@ class LaunchedKubernetesContainer(interfaces.LaunchedContainer):
             # HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"PodLogOptions \"task-pod-xxxxx\" is invalid: stream: Forbidden: may not be specified","reason":"Invalid","details":{"name":"task-pod-xxxxx","kind":"PodLogOptions","causes":[{"reason":"FieldValueForbidden","message":"Forbidden: may not be specified","field":"stream"}]},"code":422}
             # stream="All",
             _request_timeout=launcher._request_timeout,
+            _preload_content=False,
         )
+        try:
+            log = response.data.decode("utf-8", errors="replace")
+            if "\N{REPLACEMENT CHARACTER}" in log:
+                _logger.warning(
+                    f"Pod log for {self._pod_name} contained invalid UTF-8 bytes; substituted replacement characters."
+                )
+            return log
+        finally:
+            response.release_conn()
 
     def upload_log(self):
         launcher = self._get_launcher()
@@ -1490,14 +1501,26 @@ class LaunchedKubernetesJob(interfaces.LaunchedContainer):
         launcher = self._get_launcher()
         core_api_client = k8s_client_lib.CoreV1Api(api_client=launcher._api_client)
         try:
-            log = core_api_client.read_namespaced_pod_log(
+            # _preload_content=False bypasses the kubernetes client's strict .decode('utf8'),
+            # which would raise UnicodeDecodeError on torn multi-byte chars in pod logs (e.g.
+            # tqdm block glyphs split across concurrent-writer chunk boundaries).
+            response = core_api_client.read_namespaced_pod_log(
                 name=pod_name,
                 namespace=self._namespace,
                 container=_MAIN_CONTAINER_NAME,
                 timestamps=True,
                 _request_timeout=launcher._request_timeout,
+                _preload_content=False,
             )
-            return log
+            try:
+                log = response.data.decode("utf-8", errors="replace")
+                if "\N{REPLACEMENT CHARACTER}" in log:
+                    _logger.warning(
+                        f"Pod log for {pod_name} contained invalid UTF-8 bytes; substituted replacement characters."
+                    )
+                return log
+            finally:
+                response.release_conn()
         except kubernetes.client.exceptions.ApiException as ex:
             if ex.reason == "Bad Request":
                 # Kubernetes client raises kubernetes.client.exceptions.ApiException when Pod is still in PodInitializing phase


morgan-wowk:
* **#277** <a href="https://app.graphite.com/github/pr/TangleML/tangle/277?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/277?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

morgan-wowk:
🤖  **Bugsnag impact — 3.96K occurrences in the last 7 days**

This PR directly resolves the top recurring `UnicodeDecodeError` error group in Bugsnag (3,960 occurrences over the past 7 days). Two distinct variants are hitting the same strict `.decode('utf8')` call in the Kubernetes client:

1. `invalid continuation byte` — 64 worker processes writing tqdm block glyphs (`█▉▊▋▌▍▎▏`, 3-byte UTF-8: `E2 96 8x`) to the same stderr fd interleave their byte streams, leaving orphaned continuation bytes in the raw K8s API response.

2. `unexpected end of data` at `0xe2` — the K8s API response is truncated mid-sequence at a size boundary (observed around 5.1 MB), cutting a 3-byte glyph after its first byte.

Both variants throw before the run has any real failure, bubble through the retry wrapper in `internal_process_one_running_execution`, and mark otherwise-healthy training runs as `SYSTEM_ERROR` — causing the downstream Upload steps to be skipped.

**How this fixes it:** switching to `_preload_content=False` bypasses the client's strict decode entirely. We then call `.decode("utf-8", errors="replace")` ourselves, which substitutes `U+FFFD` for any malformed byte sequence — covering both error classes above. The change is applied to both the orchestrator log-upload path (`_get_log_by_pod_key`) and the live log-read path called by the API when a user views logs for a running execution (`LaunchedKubernetesContainer.get_log`). A `_logger.warning` fires whenever substitution occurs, so future occurrences remain observable without needing a repro.

cloud_pipelines_backend/launchers/kubernetes_launchers.py
```diff
@@ -931,7 +932,18 @@ def get_log(self) -> str:
             # HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"PodLogOptions \"task-pod-xxxxx\" is invalid: stream: Forbidden: may not be specified","reason":"Invalid","details":{"name":"task-pod-xxxxx","kind":"PodLogOptions","causes":[{"reason":"FieldValueForbidden","message":"Forbidden: may not be specified","field":"stream"}]},"code":422}
             # stream="All",
             _request_timeout=launcher._request_timeout,
+            _preload_content=False,
         )
+        try:
+            log = response.data.decode("utf-8", errors="replace")
+            if "�" in log:
```
Ark-kun: Let's keep the source code mostly ASCII. Let's use "\\N{REPLACEMENT CHARACTER}" instead of "�".

cloud_pipelines_backend/launchers/kubernetes_launchers.py
```diff
@@ -931,7 +932,18 @@ def get_log(self) -> str:
...
+                _logger.warning(
+                    "Pod log for %s contained invalid UTF-8 bytes; substituted replacement characters.",
```
Ark-kun: f-string, maybe?

Ark-kun (approved):

Ark-kun:
Thank you for the investigation and the fix. Approved. See the small comments.

Ark-kun:
Please add "Fixes:" line with links to the GitHub issues this PR closes

cloud_pipelines_backend/launchers/kubernetes_launchers.py
morgan-wowk: f-string sounds good.  
  
Changed

cloud_pipelines_backend/launchers/kubernetes_launchers.py
morgan-wowk: Nice suggestion. Changed.  
  
Thanks for reviewing.

morgan-wowk:
> Please add "Fixes:" line with links to the GitHub issues this PR closes

​Thanks for the review. Added the reference to the GitHub issue.

morgan-wowk:
Smoke tested locally :white_check_mark:  
Smoke tested on known upstream services / wrappers :white_check_mark:

PR #288, opened 2026-06-26
feat: Executions - API - Expose container status history on the details endpoint

morgan-wowk:
## What

Adds a `status_history` field to `GetExecutionInfoResponse` (served by **GET `/api/executions/{id}/details`**).

The status-transition instrumentation in `orchestrator_sql.py` already records, on every container-execution status change (deduped on consecutive identical statuses), an entry `{status, first_observed_at}` into `ExecutionNode.extra_data["container_execution_status_history"]`. That data was **DB-only** — no endpoint exposed it. This surfaces it on the details endpoint (which is keyed on the `ExecutionNode`, the same entity that owns `extra_data`, and is available regardless of container launch).

## Why

Clients can now compute **true time-in-status** for a task. The list is ordered and its **last entry is the node's current status**, so `status_history[-1].first_observed_at` is exactly when the node entered that status — including pre-launch states like `QUEUED`/`PENDING` that have no `ContainerExecution` row (so `/container_state` can't help). 

## Details

- New `StatusHistoryEntry` dataclass: `{status: str, first_observed_at: datetime}`.
- `first_observed_at` is parsed from the stored ISO-8601 string to a `datetime`.
- Malformed/partial entries (missing either key) are skipped; absence yields `null` rather than `[]`.
- Purely additive, optional field — no change to existing fields or callers (`kw_only` dataclass with a default).

## Tests

`tests/test_execution_nodes_api_service.py` — new `TestGetExecutionInfo`:
- returns history parsed and in order; last entry's timestamp is a tz-aware `datetime`
- `None` when no history present
- skips malformed entries

`pytest tests/test_execution_nodes_api_service.py` → 7 passed. `black` clean.

### DIFF
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -502,6 +502,12 @@ def _construct_constant_artifact_node_and_add_to_session(
 # ============
 
 
+@dataclasses.dataclass
+class ExecutionStatusHistoryEntry:
+    status: str
+    first_observed_at: datetime.datetime
+
+
 @dataclasses.dataclass(kw_only=True)
 class GetExecutionInfoResponse:
     id: bts.IdType
@@ -512,6 +518,10 @@ class GetExecutionInfoResponse:
     # ancestor_breadcrumbs: list[tuple[str, str]]
     input_artifacts: dict[str, "ArtifactNodeIdResponse"] | None = None
     output_artifacts: dict[str, "ArtifactNodeIdResponse"] | None = None
+    # Ordered history of container-execution status transitions for this node,
+    # sourced from `ExecutionNode.extra_data`. The last entry corresponds to the
+    # current status, so its `first_observed_at` is when the node entered it.
+    status_history: list[ExecutionStatusHistoryEntry] | None = None
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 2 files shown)


yuechao-qin (approved):

cloud_pipelines_backend/api_server_sql.py
```diff
@@ -502,6 +502,12 @@ def _construct_constant_artifact_node_and_add_to_session(
 # ============
 
 
+@dataclasses.dataclass
+class StatusHistoryEntry:
```
Ark-kun: Maybe, `ExecutionStatusHistoryEntry`?

Ark-kun (approved):
Thank you, Morgan.

cloud_pipelines_backend/api_server_sql.py
morgan-wowk: I like that better. Thanks for the suggestion

morgan-wowk:
Tested locally using:

```
curl -s "http://localhost:8000/api/executions/019f4d83037bf56c8167/details" \
| python3 -c '
import sys, json
d = json.load(sys.stdin)
print("id:                ", d.get("id"))
print("component:         ", (d.get("task_spec") or {}).get("componentRef", {}).get("name"))
print("parent_execution:  ", d.get("parent_execution_id"))
print("pipeline_run_id:   ", d.get("pipeline_run_id"))
print("child_task_ids:    ", d.get("child_task_execution_ids"))
print("status_history:    ", json.dumps(d.get("status_history"), indent=2))
' 2>&1
```

Working as expected.

PR #292, opened 2026-07-06
feat: Add auto-termination of stuck PENDING container executions after configured timeout

yuechao-qin:
## Summary

Closes https://github.com/Shopify/oasis-backend/issues/422

- Adds a configurable `pending_execution_timeout` to the orchestrator that auto-terminates container executions stuck in PENDING state beyond the threshold (default: disabled, oasis-backend sets 12h)
- Terminates the K8s pod, uploads logs best-effort, marks the execution as SYSTEM_ERROR with a user-visible error message, and skips downstream nodes
- Prevents zombie PENDING pods from degrading orchestrator throughput by consuming polling cycles indefinitely

## Changes

- `orchestrator_sql.py`: New `_terminate_timed_out_pending_execution` method + `_format_duration` helper
- `orchestrator_main_oasis.py`: Explicitly passes `pending_execution_timeout=timedelta(hours=12)`
- `tests/test_orchestrator_sql.py`: New test file covering timeout termination, no-op when not expired, disabled mode, and RUNNING not affected

## Context

Closes https://github.com/Shopify/oasis-backend/issues/422

When a K8s pod fails to schedule (e.g., resource exhaustion, node pool issues), the container execution remains PENDING forever. The orchestrator polls it every sweep cycle but never takes action, wasting batch budget and delaying real executions behind it in the queue.

## Test plan

- [x] Unit tests for expired, not-expired, disabled (None), and RUNNING scenarios
- [x] Unit tests for `_format_duration` at boundary values (seconds/minutes/hours)
- [x] All 443 existing tests pass

### DIFF
--- a/cloud_pipelines_backend/orchestrator_sql.py
+++ b/cloud_pipelines_backend/orchestrator_sql.py
@@ -37,6 +37,16 @@ class OrchestratorError(RuntimeError):
     pass
 
 
+def _format_duration(*, duration: datetime.timedelta) -> str:
+    """Formats a timedelta as a human-readable string with the most appropriate unit."""
+    total_seconds = duration.total_seconds()
+    if total_seconds >= 3600:
+        return f"{total_seconds / 3600:.1f} hours"
+    if total_seconds >= 60:
+        return f"{total_seconds / 60:.1f} minutes"
+    return f"{total_seconds:.1f} seconds"
+
+
 class OrchestratorService_Sql:
     def __init__(
         self,
@@ -51,6 +61,7 @@ class OrchestratorService_Sql:
         sleep_seconds_between_queue_sweeps: float = 1.0,
         output_data_purge_duration: datetime.timedelta = None,
         *,
+        pending_execution_timeout: datetime.timedelta | None = None,
         # Internal/experimental:
         _max_queue_batch_size: int = 1,
         _max_queue_batch_duration: datetime.timedelta = datetime.timedelta(),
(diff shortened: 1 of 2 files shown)


yuechao-qin:
* **#293** <a href="https://app.graphite.com/github/pr/TangleML/tangle/293?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#292** <a href="https://app.graphite.com/github/pr/TangleML/tangle/292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

morgan-wowk:
🤖 This is a clean, well-scoped fix — really nice work. The teardown path is self-contained and easy to follow (terminate → SYSTEM_ERROR → record message → skip downstream → commit), the guards read clearly, and I like that it's opt-in per deployment with `RUNNING` left untouched (nice test for that) and `_format_duration` covering the boundaries.

On the 12h default — it's a reasonable starting point, and I think it's worth grounding in what we've actually seen rather than picking it by feel. The status-change history is already recorded in each execution's `extra_data`, so we can back the number out of real data: pull the retroactive traces for execution status transitions and look at how long historical tasks legitimately dwell in PENDING before they start (the PENDING→RUNNING gap), so the threshold clears the tail of that distribution. Worth keeping in mind that `created_at` includes Kueue queue time, so that gap is exactly what this deadline clips.

One suggestion for this first version: the termination message is generic ("stuck in PENDING for X"), so from the SYSTEM_ERROR alone an operator can't tell a gcsfuse mount wedge (`CreateContainerConfigError`) from an image-pull stall or an unschedulable pod. Since we're already holding the pod status here, folding the main container's `waiting.reason`/`message` into the message would make each termination self-explanatory in the UI. There's a small property that does exactly that pull in [#289](https://github.com/TangleML/tangle/pull/289) if it's handy as a reference — feel free to lift it or leave it. Even just appending the reason string would go a long way.

Ark-kun:
Thank you for the effort, Yue Chao.  
The generic idea is straightforward, but I think it might be a bit problematic in practice:  
For "irrecoverably stuck" executions, 12 hours is too long. We'd want someting like 5-10 minutes. However for workloads that are "waiting for GPU quota" several hours can be a norm.  
One idea is to move the stuck container detection to Kubernetes launcher and make it more specific (i.e. explicitly detecting certain reasons for container being stuck). Then it can raise LauncherError. Or switch the LaunchedContainer.status to ERROR. This is a bit hacky, but we can trigger the stuck pod check when refreshing the launched container and then expose the stuck status via LaunchedContainer.status.

PR #297, opened 2026-07-16
chore: API Server - Return 409 instead of 500 when container execution not yet available

### LINKED ISSUE #296: container_state/container_log/stream endpoints return 500 (Bugsnag noise) when container execution not yet available
## Summary
RuntimeError: Execution with id='...' does not have container execution information.
surfaces in Bugsnag as an uncaught 500. This is an expected transient state (execution
PENDING, pod/container not yet created), not a server fault. Should return a non-error
HTTP status and not notify Bugsnag.

## Root cause
An ExecutionNode row exists before its ContainerExecution row. The orchestrator creates
and links the ContainerExecution only after launch (orchestrator_sql.py: created ~L642,
linked execution.container_execution = container_execution ~L674). Meanwhile
ExecutionNode.container_execution_status can already be PENDING. Window:
status == PENDING but execution.container_execution is None.

Three endpoints raise in that window (cloud_pipelines_backend/api_server_sql.py):
- GET /api/executions/{id}/container_state       → get_container_execution_state (~L765) → RuntimeError
- GET /api/executions/{id}/container_log          → get_container_execution_log (~L894) → RuntimeError
- GET /api/executions/{id}/stream_container_log   → stream_container_execution_log (~L965) → ApiServiceError

api_router.py has handlers only for ItemNotFoundError→404, PermissionError→403,
ItemAlreadyExistsError→409, ApiValidationError→422, NotImplementedError→501. RuntimeError/
ApiServiceError are uncaught → FastAPI default 500 → Bugsnag.

## Impact
Recurring Bugsnag noise. Execution stuck in PENDING (no pod, various causes) → repeated
500s as the UI polls. Real "execution does not exist" 404s buried under expected-state noise.

## Suggested solution
1. New exception e.g. ContainerExecutionNotReadyError in errors.py, distinct from
   ItemNotFoundError (which means the execution itself is absent).
2. Raise it at all three sites when the ExecutionNode exists but container_execution is None
   and status is pre-launch/pending. Unify the current RuntimeError + ApiServiceError to it.
3. Add an api_router.py handler → 409 Conflict (see below). Machine-readable body e.g.
   {"reason": "container_not_ready", "execution_status": "<status>"}. Do NOT notify Bugsnag.[8:31 AM]## HTTP status recommendation
- 409 Conflict (recommended): resource exists, current state (PENDING, no pod) conflicts
  with the request; retryable; distinguishes from 404 (execution absent) and 500 (fault).
- 404: acceptable but weaker — conflates with "execution does not exist" (already 404) and
  "wrong URL"; client can't tell "not yet" from "never".
- 200 with explicit state (for /container_state only): return status: PENDING with container
  fields null. Needs a nullable response-schema change; richest long-term answer for the poll loop.

Precedent: get_container_execution_log already returns a graceful response (not raise) for
SYSTEM_ERROR with no container row (~L878). Extend that "no row is a valid answer" pattern.

## Acceptance
- container_state/container_log/stream for a PENDING execution with no container row → 409
  (or chosen code), not 500.
- No Bugsnag event on that path.
- Genuine "execution does not exist" still returns 404

morgan-wowk:
Closes #296

## Problem

Three read endpoints return **500 + a Bugsnag notification** when an `ExecutionNode` exists but has no linked `ContainerExecution` row:

- `GET /api/executions/{id}/container_state`
- `GET /api/executions/{id}/container_log`
- `GET /api/executions/{id}/stream_container_log`

They gate on the relationship (`if not execution.container_execution`) and raised `RuntimeError`/`ApiServiceError`, which had no handler in `api_router.py`, so they fell through to the catch-all `Exception` handler → 500 + Bugsnag.

A node can legitimately have no container row in several committed states — `container_execution_status` and `container_execution` are **independent columns** with no invariant coupling them:

- **Graph / sub-pipeline nodes** — never launch a container; their status stays `None` permanently. The frontend polls these (its fetch fires on `!status`), so this is the dominant, guaranteed source of the noise.
- **Pre-launch container nodes** — `QUEUED` / `WAITING_FOR_UPSTREAM` before the orchestrator launches; hit by direct API callers (tangle-deploy, operator scripts).
- **Admin status override** (`PUT /api/admin/execution_node/{id}/status`) — can set any status, including `PENDING`, without a container link.

In all of these the condition is expected, not a server fault. The recurring 500s also buried genuine `execution does not exist` 404s under expected-state noise. (Note: the orchestrator's normal launch/reuse paths do **not** produce this — they write status and link atomically in one transaction; full analysis in the investigation doc.)

## Changes

- **`errors.py`** — new `ContainerExecutionNotReadyError` carrying `execution_status`, distinct from `ItemNotFoundError` (which means the execution itself is absent).
- **`api_server_sql.py`** — raise it at all three sites when the node exists but `container_execution` is `None`, unifying the previous `RuntimeError` + `ApiServiceError`. The existing graceful `SYSTEM_ERROR`-with-no-row path is unchanged.
- **`api_router.py`** — handler mapping it to **409 Conflict** with a machine-readable body: `{"reason": "container_not_ready", "execution_status": "<status-or-null>", "message": "..."}`.

The fix keys off the container link (not the status), so it covers every case above regardless of the node's status.

## System impact

The dedicated `ContainerExecutionNotReadyError` handler is more specific than the catch-all `Exception` handler in `api_server_main.py`, so it takes precedence and **Bugsnag is not notified** on this path. Genuine `execution does not exist` still returns **404**.

| | Before | After |
|---|---|---|
| Node exists, no container row (graph node, pre-launch, etc.) | 500 + Bugsnag event | 409, no Bugsnag |
| Execution absent | 404 | 404 (unchanged) |

## Tests

Two focused service-level tests: a node with no container row raises `ContainerExecutionNotReadyError` (with `execution_status` populated), and a genuinely absent execution still raises `ItemNotFoundError`.

### DIFF
--- a/cloud_pipelines_backend/api_router.py
+++ b/cloud_pipelines_backend/api_router.py
@@ -108,6 +108,19 @@ def _setup_routes_internal(
             response.headers["x-tangle-request-id"] = request_id
         return response
 
+    @app.exception_handler(errors.ContainerExecutionNotReadyError)
+    def handle_container_execution_not_ready_error(
+        request: fastapi.Request, exc: errors.ContainerExecutionNotReadyError
+    ):
+        return fastapi.responses.JSONResponse(
+            status_code=fastapi.status.HTTP_409_CONFLICT,
+            content={
+                "reason": "container_not_ready",
+                "execution_status": exc.execution_status,
+                "message": str(exc),
+            },
+        )
+
     @app.exception_handler(errors.PermissionError)
     def handle_permission_error(request: fastapi.Request, exc: errors.PermissionError):
         response = fastapi.responses.JSONResponse(
--- a/cloud_pipelines_backend/api_server_sql.py
+++ b/cloud_pipelines_backend/api_server_sql.py
@@ -761,8 +761,9 @@ class ExecutionNodesApiService_Sql:
             raise errors.ItemNotFoundError(f"Execution with {id=} does not exist.")
         container_execution = execution.container_execution
         if not container_execution:
-            raise RuntimeError(
-                f"Execution with {id=} does not have container execution information."
+            raise errors.ContainerExecutionNotReadyError(
+                execution_node_id=id,
+                execution_status=execution.container_execution_status,
             )
 
         if include_execution_nodes_linked_to_same_container_execution:
@@ -890,8 +891,9 @@ class ExecutionNodesApiService_Sql:
                     system_error_exception_full=system_error_exception_full,
                     orchestration_error_message=orchestration_error_message,
                 )
-            raise RuntimeError(
-                f"Execution with {id=} does not have container execution information."
+            raise errors.ContainerExecutionNotReadyError(
+                execution_node_id=id,
+                execution_status=execution.container_execution_status,
             )
         log_text: str | None = None
         if container_execution.status in (
@@ -961,8 +963,9 @@ class ExecutionNodesApiService_Sql:
             )
         container_execution = execution.container_execution
         if not container_execution:
-            raise ApiServiceError(
-                "Execution does not have container execution information."
+            raise errors.ContainerExecutionNotReadyError(
+                execution_node_id=execution_id,
+                execution_status=execution.container_execution_status,
             )
         if not container_execution.launcher_data:
             raise ApiServiceError(
--- a/cloud_pipelines_backend/errors.py
+++ b/cloud_pipelines_backend/errors.py
@@ -2,6 +2,18 @@ class ItemNotFoundError(Exception):
     pass
 
 
+class ContainerExecutionNotReadyError(Exception):
+    """Execution exists but its container execution is not yet available (transient)."""
+
+    def __init__(self, *, execution_node_id: str, execution_status: str | None = None):
+        self.execution_node_id = execution_node_id
+        self.execution_status = execution_status
+        super().__init__(
+            f"Execution with {execution_node_id=} does not have "
+            "container execution information."
+        )
+
+
 class ItemAlreadyExistsError(Exception):
     pass
 
--- a/tests/test_execution_nodes_api_service.py
+++ b/tests/test_execution_nodes_api_service.py
@@ -6,6 +6,7 @@ from sqlalchemy import orm
 from cloud_pipelines_backend import backend_types_sql as bts
 from cloud_pipelines_backend import component_structures as structures
 from cloud_pipelines_backend import database_ops
+from cloud_pipelines_backend import errors
 from cloud_pipelines_backend.api_server_sql import (
     ExecutionNodesApiService_Sql,
     GetExecutionInfoResponse,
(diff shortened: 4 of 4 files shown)


morgan-wowk:
* **#297** <a href="https://app.graphite.com/github/pr/TangleML/tangle/297?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/297?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

cloud_pipelines_backend/api_router.py
```diff
@@ -108,6 +111,23 @@ def handle_not_found_error(request: fastapi.Request, exc: errors.ItemNotFoundErr
             response.headers["x-tangle-request-id"] = request_id
         return response
 
+    @app.exception_handler(errors.ContainerExecutionNotReadyError)
+    def handle_container_execution_not_ready_error(
+        request: fastapi.Request, exc: errors.ContainerExecutionNotReadyError
+    ):
+        with contextual_logging.logging_context(execution_id=exc.execution_node_id):
+            _logger.info(
+                "Container execution not ready (status=%s)", exc.execution_status
```
morgan-wowk: nit: We can bake execution status into `logging_context` above

Ark-kun:
>During that window the node's container_execution_status can already be PENDING while execution.container_execution is still None.

Are you sure this can happen? At first glance this looks like it would be a bug.

Ark-kun:
>handler mapping it to 409 Conflict 

Would 404 be more appropriate? The container sub-resource does not exist (yet).

cloud_pipelines_backend/api_router.py
```diff
@@ -108,6 +111,23 @@ def handle_not_found_error(request: fastapi.Request, exc: errors.ItemNotFoundErr
             response.headers["x-tangle-request-id"] = request_id
         return response
 
+    @app.exception_handler(errors.ContainerExecutionNotReadyError)
+    def handle_container_execution_not_ready_error(
+        request: fastapi.Request, exc: errors.ContainerExecutionNotReadyError
+    ):
+        with contextual_logging.logging_context(execution_id=exc.execution_node_id):
+            _logger.info(
```
Ark-kun: Not sure we need to log this. This is a common occurrence.

morgan-wowk:
> > handler mapping it to 409 Conflict
>
> Would 404 be more appropriate? The container sub-resource does not exist (yet).

​I was thinking 404 but it can be confused with the execution entity / DB record not being found. 409 is an alternative that could be used to say the route was valid, and found a record, but could not produce the intended result

morgan-wowk:
> > During that window the node's container_execution_status can already be PENDING while execution.container_execution is still None.
>
> Are you sure this can happen? At first glance this looks like it would be a bug.

​I shall verify

Ark-kun (approved):
Approved with comments.

morgan-wowk:
> > During that window the node's container_execution_status can already be PENDING while execution.container_execution is still None.
>
> Are you sure this can happen? At first glance this looks like it would be a bug.

​This has been disproven and the statement will be removed from this PR description. Tangle UI actively polls for container execution information for subgraph nodes when said data will never exist. An issue has been created in Tangle UI: https://github.com/TangleML/tangle-ui/issues/2539

cloud_pipelines_backend/api_router.py
morgan-wowk: Thank you for letting me know. I was on the fence about the log.

morgan-wowk: Good call — removed the log line (and the now-unused `_logger`/`logging` import). It was a common, expected occurrence and the 409 response is self-describing.

morgan-wowk:
- Smoke tested locally with OSS environment
- Smoke tested with known upstream consumers

PR #298, opened 2026-07-16
Recover saved Kubernetes terminal state

nateGeorge:
Tangle now finishes from already-saved container state after a restart instead of trying to read a deleted Kubernetes resource. For Jobs, a finished parent result remains valid when child Pod details or logs are no longer available.

Draft follow-up #299 adds an opt-in single-node Job mode so Kubernetes can retain terminal state after a completed task Pod is removed.

### Testing

- `.venv/bin/black --check cloud_pipelines_backend/launchers/interfaces.py cloud_pipelines_backend/launchers/kubernetes_launchers.py cloud_pipelines_backend/orchestrator_sql.py tests/test_kubernetes_terminal_recovery.py` (passed)
- `uv run --frozen --with-editable . pytest -q tests/test_kubernetes_terminal_recovery.py` (34 passed)
- `uv run --frozen --with-editable . pytest -q` (471 passed, 16 warnings)
- `git diff --check` (passed)

### DIFF
--- a/cloud_pipelines_backend/launchers/interfaces.py
+++ b/cloud_pipelines_backend/launchers/interfaces.py
@@ -71,6 +71,25 @@ class ContainerStatus(str, enum.Enum):
     ERROR = "ERROR"
 
 
+class LauncherResourceNotFound(LauncherError):
+    def __init__(
+        self,
+        *,
+        kind: str,
+        namespace: str,
+        name: str,
+        cached_status: ContainerStatus,
+    ):
+        self.kind = kind
+        self.namespace = namespace
+        self.name = name
+        self.cached_status = cached_status
+        super().__init__(
+            f"Kubernetes {kind} {namespace}/{name} was not found while refreshing "
+            f"cached {cached_status.value} state."
+        )
+
+
 class LaunchedContainer(abc.ABC):
 
     # @classmethod
--- a/cloud_pipelines_backend/launchers/kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/kubernetes_launchers.py
@@ -6,6 +6,7 @@ import json
 import logging
 import os
 import pathlib
+import time
 import typing
 from typing import Any, Optional
 
@@ -71,6 +72,21 @@ _MULTI_NODE_NODE_INDEX_ENV_VAR_NAME = "_TANGLE_MULTI_NODE_NODE_INDEX"
 
 
 _T = typing.TypeVar("_T")
+
+_KUBERNETES_DIAGNOSTIC_MAX_ATTEMPTS = 3
+_KUBERNETES_DIAGNOSTIC_RETRY_WAIT_SECONDS = 0.1
+
+
+def _retry_kubernetes_diagnostic_call(func: typing.Callable[[], _T]) -> _T:
+    for attempt in range(_KUBERNETES_DIAGNOSTIC_MAX_ATTEMPTS):
+        try:
+            return func()
+        except Exception:
+            if attempt == _KUBERNETES_DIAGNOSTIC_MAX_ATTEMPTS - 1:
+                raise
+            time.sleep(_KUBERNETES_DIAGNOSTIC_RETRY_WAIT_SECONDS)
+    raise RuntimeError("Unreachable")
+
 
 _CONTAINER_FILE_NAME = "data"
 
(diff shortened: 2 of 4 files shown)


cloud_pipelines_backend/launchers/interfaces.py
```diff
@@ -71,6 +71,25 @@ class ContainerStatus(str, enum.Enum):
     ERROR = "ERROR"
 
 
+class LauncherResourceNotFound(LauncherError):
```
Ark-kun, line 74: What's your plan for using this common exception in non-Kubernetes launchers?

Ark-kun, line 74: Where is this exception used (caught)?

cloud_pipelines_backend/launchers/kubernetes_launchers.py
```diff
@@ -1449,28 +1483,50 @@ def from_dict(
...
+                f"Unable to collect child Pod diagnostics for terminal Job {self._job_name} in namespace {self._namespace}."
+            )
+            return new_launched_container
+
         pod_map: dict[str, k8s_client_lib.V1Pod] = {}
         if job.spec.completion_mode == "Indexed":
-            # There can be multiple pods for each index.
-            # This can happen when Job gets suspended and resumed multiple times and Pods can get stuck in "Terminating" phase.
-            # In this case we want to get the latest Pods.
-            # Alternatively, we could store all pods, but this can complicate the routines that determine status and get logs.
```
Ark-kun, line 1473: Can you please explain what was your intent when making this change?

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -212,8 +212,13 @@ def internal_process_running_executions_queue(self, session: orm.Session):
...
 
-                    # Mark our ExecutionNode as SYSTEM_ERROR
-                    execution_nodes = running_container_execution.execution_nodes
+                    # Mark our active ExecutionNodes as SYSTEM_ERROR.
+                    execution_nodes = [
+                        execution_node
+                        for execution_node in running_container_execution.execution_nodes
+                        if execution_node.container_execution_status
+                        != bts.ContainerExecutionStatus.CANCELLED
+                    ]
```
Ark-kun, line 221: Can you please explain this change?

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -689,70 +694,77 @@ def internal_process_one_running_execution(
...
-
-        # Asking the launcher to refresh the container state.
-        reloaded_launched_container: launcher_interfaces.LaunchedContainer = (
-            self._launcher.get_refreshed_launched_container_from_dict(launcher_data)
+        cached_snapshot_is_terminal = launched_container.status in (
+            launcher_interfaces.ContainerStatus.SUCCEEDED,
+            launcher_interfaces.ContainerStatus.FAILED,
         )
+
+        if not cached_snapshot_is_terminal:
```
Ark-kun, line 713: When is `cached_snapshot_is_terminal` true?

nateGeorge:
Closing this in favor of the existing Kubernetes Job path. Setting `tangleml.com/launchers/kubernetes/multi_node/number_of_nodes: "1"` routes a single-node task through the Job launcher, and we have verified that workaround end to end.

The underlying gap still exists for direct Pod executions: if the Pod is deleted before the orchestrator observes its terminal state, refreshing the live resource can turn completed work into `SYSTEM_ERROR`. A Job preserves terminal status, although its child Pod logs can still disappear. This PR needs broader design alignment and I do not have capacity to continue it right now.

nateGeorge:
Follow-up with a concrete reproduction and the design conclusion.

In Oasis run `019fa9d270d96dfaa226`, execution `019fa9d26bfc8370b702` launched bare Pod `task-019fa9e020b0b5f80592-fl4nb` for `score_shard_01`.

The workload did complete successfully:

- `18:06:37Z`: it wrote the final prediction batch.
- `18:06:41.029Z`: it returned `flat_prediction_table=... row_count=2359396 skipped=False`.
- `18:06:42Z`: the GCS Fuse sidecar received `SIGTERM`, consistent with the main workload having exited.
- `18:06:47Z`: the sidecar exited.
- `18:09:33Z`: during later reconciliation/log retrieval, Tangle called Kubernetes for the Pod and received `404 Not Found: pods "task-019fa9e020b0b5f80592-fl4nb" not found`.
- The exception exhausted retry handling, the execution became `SYSTEM_ERROR`, and dependent tasks were skipped even though the output table had been written.

This also clarifies why switching to a Job is useful but not a complete fix. A Job preserves durable terminal state after its child Pod disappears, but it does not preserve the deleted Pod's logs.

A thorough fix should separate terminal-state durability from log durability:

1. Use Jobs for batch tasks, including single-node tasks, so reconciliation can read completion and exit status from the Job after the Pod is gone.
2. Stream container logs to durable external storage while the Pod is running; do not rely on fetching logs from a deleted Pod.
3. Persist the terminal status, exit code, and launcher snapshot before cleanup when possible.
4. Treat post-completion log retrieval as best-effort. Missing logs must not turn a known successful workload into `SYSTEM_ERROR`.
5. If both Pod and durable parent state are missing while status is still unknown, report an inconclusive infrastructure failure rather than guessing success.

For our immediate case, the existing `tangleml.com/launchers/kubernetes/multi_node/number_of_nodes: "1"` annotation already routes the task through the Kubernetes Job launcher, so we are using that instead of maintaining the proposed launcher changes in #299. I closed this PR because the recovery behavior here needs to be designed together with Job reconciliation and durable log handling rather than patched independently.

PR #299, opened 2026-07-17
Add opt-in single-node Kubernetes Jobs

nateGeorge:
Adds an opt-in NonIndexed Job mode for single-node tasks while keeping direct Pods as the default and preserving existing Indexed Job behavior for multi-node tasks.

This is a follow-up to #298 and must merge after it. Launcher defaults can select `pod` or `job`, individual tasks can override the single-node mode with an annotation, and multi-node annotations always take precedence.

### Testing

- `uv run --frozen black --check --fast cloud_pipelines_backend/launchers/kubernetes_launchers.py cloud_pipelines_backend/launchers/google_kubernetes_launchers.py tests/test_kubernetes_launchers.py`
- `PYTHONPATH=. uv run --frozen pytest -q tests/test_kubernetes_launchers.py tests/test_skypilot_launchers.py` — 35 passed
- `PYTHONPATH=. uv run --frozen pytest -q` — 449 passed, 13 warnings
- `git diff --check`

### DIFF
--- a/cloud_pipelines_backend/launchers/google_kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/google_kubernetes_launchers.py
@@ -54,7 +54,12 @@ class GoogleKubernetesEngine_UsingGoogleCloudStorage_KubernetesJobLauncher(
 class GoogleKubernetesEngine_UsingGoogleCloudStorage_KubernetesPodOrJobLauncher(
     kubernetes_launchers._KubernetesPodOrJobLauncher
 ):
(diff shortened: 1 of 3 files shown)


Ark-kun:
@nateGeorge You can already make the launcher use Kubernetes Jobs for single-node tasks by specifying "Number of nodes: 1".

nateGeorge:
Closing because the existing launcher already supports the needed behavior. As Alexey noted, adding `tangleml.com/launchers/kubernetes/multi_node/number_of_nodes: "1"` routes the task through the Kubernetes Job launcher. We have verified that workaround end to end, so a second single-node execution-mode API is not needed for the immediate use case.

PR #312, opened 2026-08-06
fix(orchestrator): optionally retry retriable container-refresh failures

morgan-wowk:
## What

When refreshing a running container's state fails, the orchestrator used to immediately mark it `SYSTEM_ERROR` and skip its whole downstream DAG — even when the failure was just the backing platform being briefly unavailable (e.g. a `500 … RST_STREAM ENHANCE_YOUR_CALM` from the Kubernetes API server). We learned nothing about the container, so throwing away the task and its DAG is wrong.

This PR lets the orchestrator **tolerate a few consecutive retriable refresh failures** before terminalizing.

- Whether a failure is retriable is the **launcher's** decision. `LauncherError` gains an `is_retriable` flag; the Kubernetes launcher sets it for 5xx while refreshing a pod/job. The orchestrator acts on the flag and no longer inspects HTTP status codes.
- Each running execution gets a budget of consecutive retriable failures (default 3). A tolerated failure is naturally sent to the back of the queue; any successful refresh resets the budget. Non-retriable errors terminalize immediately, as before.

## Behind a feature flag (off by default)

The behaviour change is gated on the `TANGLE_RETRY_CONTAINER_REFRESH_FAILURES` env var. **Disabled by default → identical to today** (any refresh failure terminalizes immediately). Enabling it turns on the retry budget.

## Scope

- Only the **refresh** is retried. The container is never relaunched, so non-idempotent programs are never re-run by this path.

### DIFF
--- a/cloud_pipelines_backend/launchers/kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/kubernetes_launchers.py
@@ -595,7 +595,12 @@ class _KubernetesPodLauncher(
         launched_container = LaunchedKubernetesContainer.from_dict(
             launched_container_dict, launcher=self
         )
-        return launched_container.get_refreshed()
+        try:
+            return launched_container.get_refreshed()
+        except kubernetes.client.exceptions.ApiException as ex:
+            raise _launcher_error_from_api_exception(
+                ex, message="Failed to refresh pod status"
+            ) from ex
 
     def deserialize_launched_container_from_dict(
         self, launched_container_dict: dict
@@ -1257,7 +1262,12 @@ class _KubernetesJobLauncher(
         launched_container = LaunchedKubernetesJob.from_dict(
             launched_container_dict, launcher=self
         )
-        return launched_container.get_refreshed()
+        try:
+            return launched_container.get_refreshed()
+        except kubernetes.client.exceptions.ApiException as ex:
+            raise _launcher_error_from_api_exception(
+                ex, message="Failed to refresh job status"
+            ) from ex
 
     def deserialize_launched_container_from_dict(
         self, launched_container_dict: dict
@@ -1848,3 +1858,20 @@ def _remove_keys_with_none_values(d: dict):
             del d[k]
         if isinstance(v, dict):
             _remove_keys_with_none_values(v)
+
+
+def _launcher_error_from_api_exception(
+    exception: kubernetes.client.exceptions.ApiException,
+    *,
+    message: str,
+) -> interfaces.LauncherError:
+    """Translate a Kubernetes API error into a launcher error.
+
+    A 5xx is retriable (the API server broke or shed load); any other status is
+    a definitive failure.
+    """
+    status = exception.status
+    is_retriable = isinstance(status, int) and 500 <= status < 600
+    return interfaces.LauncherError(
+        f"{message}: {exception!r}", is_retriable=is_retriable
+    )
--- a/cloud_pipelines_backend/orchestrator_sql.py
+++ b/cloud_pipelines_backend/orchestrator_sql.py
@@ -2,6 +2,7 @@ import copy
 import json
 import datetime
 import logging
+import os
 import time
 import traceback
 import typing
@@ -52,6 +53,7 @@ class OrchestratorService_Sql:
         output_data_purge_duration: datetime.timedelta = None,
         *,
         # Internal/experimental:
+        _max_container_execution_refresh_error_retries: int = 3,
         _max_queue_batch_size: int = 1,
         _max_queue_batch_duration: datetime.timedelta = datetime.timedelta(),
     ):
@@ -65,6 +67,10 @@ class OrchestratorService_Sql:
         self._queued_executions_queue_idle = False
         self._running_executions_queue_idle = False
         self._output_data_purge_duration = output_data_purge_duration
+        self._max_container_execution_refresh_error_retries = (
+            _max_container_execution_refresh_error_retries
+        )
+        self._container_execution_refresh_error_counts: dict[Any, int] = {}
 
         self._max_queue_batch_size = _max_queue_batch_size
         self._max_queue_batch_duration = _max_queue_batch_duration
@@ -213,6 +219,38 @@ class OrchestratorService_Sql:
                 except Exception as ex:
                     _logger.exception("Error processing running container execution")
                     session.rollback()
+
+                    is_retriable = (
+                        isinstance(ex, launcher_interfaces.LauncherError)
+                        and ex.is_retriable
+                    )
+                    error_count = (
+                        self._container_execution_refresh_error_counts.get(
+                            running_container_execution.id, 0
+                        )
+                        + 1
+                    )
+                    if (
+                        is_retriable
+                        and error_count
+                        < self._max_container_execution_refresh_error_retries
+                    ):
+                        self._container_execution_refresh_error_counts[
+                            running_container_execution.id
+                        ] = error_count
+                        _logger.warning(
+                            "Could not refresh running container execution "
+                            f"({error_count} of "
+                            f"{self._max_container_execution_refresh_error_retries}"
+                            f" consecutive retriable failures): {ex!r}. Leaving "
+                            f"it in {running_container_execution.status} and "
+                            "refreshing it on a later sweep."
+                        )
+                        return True
+
+                    self._container_execution_refresh_error_counts.pop(
+                        running_container_execution.id, None
+                    )
                     running_container_execution.status = (
                         bts.ContainerExecutionStatus.SYSTEM_ERROR
                     )
@@ -238,6 +276,10 @@ class OrchestratorService_Sql:
                             session=session, execution=execution_node
                         )
                     session.commit()
+                else:
+                    self._container_execution_refresh_error_counts.pop(
+                        running_container_execution.id, None
+                    )
                 finally:
                     duration_ms = (time.monotonic_ns() - start_timestamp) / 1_000_000
                     _logger.info(
--- a/cloud_pipelines_backend/launchers/interfaces.py
+++ b/cloud_pipelines_backend/launchers/interfaces.py
@@ -26,7 +26,15 @@ __all__ = [
 
 
 class LauncherError(RuntimeError):
-    pass
+    """Base class for errors raised by a launcher.
+
+    `is_retriable` marks a failure worth trying again (e.g. the backing platform
+    was unavailable) rather than a definitive error.
+    """
+
+    def __init__(self, *args: object, is_retriable: bool = False) -> None:
+        super().__init__(*args)
+        self.is_retriable = is_retriable
 
 
 @dataclasses.dataclass(kw_only=True)
--- a/tests/test_container_execution_refresh_retries.py
+++ b/tests/test_container_execution_refresh_retries.py
@@ -1,0 +1,265 @@
+"""Tests for the refresh-failure budget in the running-executions queue."""
+
+from typing import Any, Callable
+from unittest import mock
+
+from sqlalchemy import orm
+from sqlalchemy import sql
+
+from cloud_pipelines_backend import api_server_sql
+from cloud_pipelines_backend import backend_types_sql as bts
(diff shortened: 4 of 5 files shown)


morgan-wowk:
* **#314** <a href="https://app.graphite.com/github/pr/TangleML/tangle/314?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#313** <a href="https://app.graphite.com/github/pr/TangleML/tangle/313?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#312** <a href="https://app.graphite.com/github/pr/TangleML/tangle/312?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/312?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -205,31 +258,99 @@ def internal_process_running_executions_queue(self, session: orm.Session):
...
+                        # to the back: every other execution is observed first,
+                        # on the very next sweep, and we come back to this one
+                        # later. Nothing sleeps and nothing else is delayed.
+                        retry_at = (
+                            _get_current_time()
+                            + self._running_execution_poll_retry_delay
                         )
-                        record_system_error_exception(
-                            execution=execution_node, exception=ex
+                        running_container_execution.last_processed_at = retry_at
```
Ark-kun: Not sure this is needed. Processed executions are already sent to the back of the queue (First few lines of the internal_process_one_running_execution code).

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -33,6 +33,49 @@
...
+    while current is not None and id(current) not in seen:
+        seen.add(id(current))
+        status = getattr(current, "status", None)
+        if isinstance(status, int) and not isinstance(status, bool):
+            return status
+        current = current.__cause__ or current.__context__
+    return None
+
+
+def _is_transient_observation_failure(exception: BaseException) -> bool:
```
Ark-kun: I think these checks are too Kubernetes-specific and belong to the Kubernetes launcher. We can add `is_retriable` to `LauncherError` and then detect that in the Orchestrator.

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -50,6 +93,10 @@ def __init__(
         default_task_annotations: dict[str, Any] | None = None,
         sleep_seconds_between_queue_sweeps: float = 1.0,
         output_data_purge_duration: datetime.timedelta = None,
+        max_running_execution_poll_failures: int = 3,
+        running_execution_poll_retry_delay: datetime.timedelta = datetime.timedelta(
```
Ark-kun: I'm not sure we really need this (right now). We can just put the execution to the end of the queue (it's already done automatically).

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -33,6 +33,49 @@
...
+        seen.add(id(current))
+        status = getattr(current, "status", None)
+        if isinstance(status, int) and not isinstance(status, bool):
+            return status
+        current = current.__cause__ or current.__context__
+    return None
+
+
+def _is_transient_observation_failure(exception: BaseException) -> bool:
+    """Whether failing to observe a container is worth re-observing later.
```
Ark-kun: Nit: The "observe" terminology sounds weird here. It's not used anywhere in orchestrator and is clashing with Observability.

We "handle" queue items.

We "refresh" the launched container state.

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -65,6 +112,12 @@ def __init__(
         self._queued_executions_queue_idle = False
         self._running_executions_queue_idle = False
         self._output_data_purge_duration = output_data_purge_duration
+        self._max_running_execution_poll_failures = max_running_execution_poll_failures
+        self._running_execution_poll_retry_delay = running_execution_poll_retry_delay
+        # Number of consecutive failures observed while *polling* the status of a
+        # running container execution, keyed by `ContainerExecution.id`. Entries
+        # are dropped as soon as a poll succeeds or the execution is terminalized.
+        self._running_execution_poll_failures: dict[Any, int] = {}
```
Ark-kun: We could store this information in the extra_data of the container execution.  
A clever way to do that is to store the failure history: a list of timestamps and maybe the associated errors - good for analysis. The number of failures is then the size of history.

Pros:

- Does not disappear after backend restart
- We do not keep the whole table in memory (although the size is really small)
- Can be analyzed by querying the DB.

Cons:

- Requires a DB commit to persist.

OK to do in different PR/later.

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -205,31 +258,99 @@ def internal_process_running_executions_queue(self, session: orm.Session):
...
+                            f"re-observing after {retry_at}."
                         )
-                    # Doing an intermediate commit here because it's most important to mark the problematic node as SYSTEM_ERROR.
-                    session.commit()
-
-                    # Skip downstream executions
-                    for execution_node in execution_nodes:
-                        _mark_all_downstream_executions_as_skipped(
-                            session=session, execution=execution_node
+                    else:
```
Ark-kun: Can we avoid extra nesting by putting `return True` here?

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -50,6 +93,10 @@ def __init__(
         default_task_annotations: dict[str, Any] | None = None,
         sleep_seconds_between_queue_sweeps: float = 1.0,
         output_data_purge_duration: datetime.timedelta = None,
+        max_running_execution_poll_failures: int = 3,
```
Ark-kun: `_max_container_execution_refresh_error_retries`

Let's make it experimental for now. And let's move it to the keyword-only section (after `*`).

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -33,6 +33,49 @@
 DYNAMIC_DATA_SECRET_NAME_KEY = "name"
 
 
+# Safety valve for `OrchestratorService_Sql._running_execution_poll_failures`.
+# Entries are normally removed as soon as a poll succeeds or the execution is
+# terminalized, but an execution row that disappears while PENDING/RUNNING would
+# otherwise leak one entry in a long-lived process.
+_MAX_TRACKED_POLL_FAILURES = 10_000
```
Ark-kun: I don't think we need this limit for now.

Later we can upgrade to storing the retry info in the DB (via extra_data).

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -33,6 +33,49 @@
 DYNAMIC_DATA_SECRET_NAME_KEY = "name"
 
 
+# Safety valve for `OrchestratorService_Sql._running_execution_poll_failures`.
+# Entries are normally removed as soon as a poll succeeds or the execution is
+# terminalized, but an execution row that disappears while PENDING/RUNNING would
```
Ark-kun: ...terminalized...?

cloud_pipelines_backend/launchers/kubernetes_launchers.py
```diff
@@ -32,6 +32,23 @@
 _MAIN_CONTAINER_NAME = "main"
 
 
+def _launcher_error_from_api_exception(
```
Ark-kun: Nit: Let's move this utility function to the end of the file.

cloud_pipelines_backend/orchestrator_sql.py
```diff
@@ -32,6 +33,10 @@
 DYNAMIC_DATA_SECRET_KEY = "secret"
 DYNAMIC_DATA_SECRET_NAME_KEY = "name"
 
+_RETRY_CONTAINER_REFRESH_FAILURES = os.environ.get(
+    "TANGLE_RETRY_CONTAINER_REFRESH_FAILURES", ""
+).lower() in ("1", "true", "yes")
+
```
Ark-kun: It's OK not to gate. The feature is strtaightforward enough and easy to disable.

Ark-kun (approved):

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Agreed — there's no manual `last_processed_at` reordering anymore. The retry path just leaves the execution in place and returns; the existing ordering by `last_processed_at` already sends a just-processed execution to the back, so nothing else is delayed (covered by `test_failing_execution_does_not_hold_up_the_queue`).

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Done. `LauncherError` now carries `is_retriable`; the Kubernetes launcher sets it for 5xx in `_launcher_error_from_api_exception`, and the orchestrator only checks `ex.is_retriable` — no status codes in the orchestrator.

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Removed the retry-delay parameter — relying on the automatic back-of-queue ordering as you suggested.

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Renamed to "refresh" throughout (the warning log, the counters, and the tests).

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Agreed — leaving the DB/`extra_data` failure-history for a follow-up. The in-memory counter is dropped on success/terminalization, and I removed the unbounded-growth safety valve here per the other comment.

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Done — early `return True` on the retry path, no extra nesting.

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Done — renamed to `_max_container_execution_refresh_error_retries` and moved into the keyword-only `Internal/experimental` block.

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Removed `_MAX_TRACKED_POLL_FAILURES`.

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: That constant and its comment are gone now (safety valve removed). "Terminalize" only remains in test names, meaning "move to a terminal status" — happy to rename if you'd prefer.

cloud_pipelines_backend/launchers/kubernetes_launchers.py
morgan-wowk: Done — moved `_launcher_error_from_api_exception` to the bottom of the file.

cloud_pipelines_backend/orchestrator_sql.py
morgan-wowk: Removed the `TANGLE_RETRY_CONTAINER_REFRESH_FAILURES` gate — the retry is on by default now, and as you say it's easy to disable by reverting. Dropped the disabled-flag test with it.

morgan-wowk:
### Merge activity

* **Aug 11, 11:45 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/312).
* **Aug 11, 11:45 PM UTC**: @morgan-wowk merged this pull request with [Graphite](https://app.graphite.com/github/pr/TangleML/tangle/312).

PR #316, opened 2026-08-07
fix: Orchestrator - Fixed executions remaining stuck in WAITING_FOR_UPSTREAM status forever when an upstream queued execution node failed to launch with SYSTEM_ERROR

### LINKED ISSUE #315: bug: Queued-execution failure strands the downstream subtree in `WAITING_FOR_UPSTREAM` forever
## What happens

When `internal_process_one_queued_execution` raises, the handler in
`OrchestratorService_Sql.internal_process_queued_executions_queue`
(`cloud_pipelines_backend/orchestrator_sql.py`) marks the failed node
`SYSTEM_ERROR` but does not mark its downstream nodes `SKIPPED`.

Nothing ever wakes those nodes: the queued handler selects only
`UNINITIALIZED`/`QUEUED` and the running handler selects on
`ContainerExecution.status`, so no queue picks up `WAITING_FOR_UPSTREAM`. The
whole downstream subtree is stranded permanently — the run never reaches a
terminal state, and the non-terminal node gauge climbs and never comes back
down.

Every other failure path (conditional skip, container failure, upstream
`SYSTEM_ERROR` during argument resolution) already calls
`_mark_all_downstream_executions_as_skipped`. This one path is the outlier.

## Reproduce

Any exception inside processing triggers it. Smallest case: a two-task graph
where `downstream` consumes `upstream`'s output and `upstream` has
`isEnabled: "maybe"` — the conditional-execution check raises because the value
resolves to neither `"true"` nor `"false"`.

Result: `upstream` is `SYSTEM_ERROR`, `downstream` stays `WAITING_FOR_UPSTREAM`
indefinitely. Expected: `downstream` is `SKIPPED`.

In production this was triggered by a burst of secret-resolution failures
(a `DynamicDataArgument` naming a secret that does not exist).

## Fix

Call `_mark_all_downstream_executions_as_skipped` in the handler, **after**
committing `SYSTEM_ERROR`. The ordering matters: committing the status first
makes it durable even if the traversal raises. Committing afterwards would let
the rollback revert the node to `QUEUED`, so the queue would hand it back and it
would fail identically on every sweep.

## Also needed

The fix stops new orphans but does not repair rows already stranded. Existing
ones need a one-off cleanup that marks the stranded subtree of an already-failed
run `SKIPPED`. It must only touch runs where nothing is still moving — a
`SYSTEM_ERROR` in one branch is not license to skip a run with `RUNNING`,
`QUEUED`, `PENDING`, `CANCELLING`, or `UNINITIALIZED` work elsewhere.

yuechao-qin:
Closes https://github.com/TangleML/tangle/issues/315

### DIFF
--- a/cloud_pipelines_backend/orchestrator_sql.py
+++ b/cloud_pipelines_backend/orchestrator_sql.py
@@ -147,6 +147,14 @@ class OrchestratorService_Sql:
                         execution=queued_execution, exception=ex
                     )
                     session.commit()
+                    # Skipping downstream happens after the commit above, so that
+                    # the SYSTEM_ERROR status is durable even if this traversal
+                    # fails. Otherwise the execution would stay QUEUED and be
+                    # reprocessed on every sweep.
+                    _mark_all_downstream_executions_as_skipped(
+                        session=session, execution=queued_execution
+                    )
+                    session.commit()
                 finally:
                     duration_ms = (time.monotonic_ns() - start_timestamp) / 1_000_000
                     _logger.info(
--- a/tests/test_orchestrator_sql.py
+++ b/tests/test_orchestrator_sql.py
@@ -1,0 +1,330 @@
+"""Tests for ``orchestrator_sql``.
+"""
+
+from typing import Callable
+from unittest import mock
+
+import pytest
+from sqlalchemy import orm
+from sqlalchemy import sql
+
+from cloud_pipelines_backend import api_server_sql
+from cloud_pipelines_backend import backend_types_sql as bts
+from cloud_pipelines_backend import component_structures as structures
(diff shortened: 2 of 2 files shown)


yuechao-qin:
* **#316** <a href="https://app.graphite.com/github/pr/TangleML/tangle/316?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/316?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Ark-kun (approved):

PR #317, opened 2026-08-07
fix(kubernetes): hold task pods until their terminal status is observed

morgan-wowk:
The orchestrator can only learn how a task ended by reading the pod object:
`LaunchedKubernetesContainer.status` reads `pod.status.phase`, exit code and
timings come off the same object, and logs are fetched from the live pod at the
moment a terminal status is first seen. Nothing is streamed or persisted before
then. So if the pod is deleted between the container finishing and the next
poll, a task that succeeded is recorded as SYSTEM_ERROR, its downstream DAG is
skipped, and its logs are gone for good.

That window is not hypothetical and it is not ours to widen. On the cluster
these tasks run on, terminated pods are collected by a count-limited sweep in
pod-age order: the measured lag between a pod finishing and being deleted moved
between ~20 and ~200 minutes over a single day, and a pod older than the current
frontier is removed within seconds of terminating (one observed task ran for 634
minutes and its pod was gone in the same minute it succeeded). Node scale-down
takes the pods with it just as fast. The orchestrator re-polls a given pod once every `active_executions / ~2.2`
seconds. Its running-executions queue is a single round-robin over every
non-terminal execution ordered by `last_processed_at`: three polls per sweep,
a 1s sleep between sweeps, ~2.2 pod polls/sec in aggregate. Measured in
production at 147 active executions, the gap between two polls of the same pod
was 65s p50 and 60-76s across 453 samples. It scales linearly with fleet
concurrency, and active executions ranged 34 to 277 over the last week, so the
window is near two minutes at peak -- widest exactly when pod churn and node
scale-down are highest. The orchestrator loses both races, and the
garbage-collection threshold is a control-plane flag that a managed Kubernetes
offering does not expose.

Close the window from this side instead. When the hold is enabled, the pod
launcher stamps every pod it creates with a
`cloud-pipelines.net/status-observed` finalizer. A delete then only stamps
`deletionTimestamp`: the pod stays in `Terminating`, fully readable, until we
remove the finalizer. This does not retain pods for longer in the normal case;
it converts a race into a handshake, and the pod is released as soon as its
outcome has been recorded.

The hold is the easy half. The release is the half that has to be exhaustive,
because a pod that is held and never released sits in `Terminating` for ever,
consuming etcd space and namespace pod quota and eating the cluster's pod-GC
budget. Every path that terminalizes an execution now releases:

- a terminal status observed by the poll loop, after status, exit code, logs
and output artifacts are all committed;
- cancellation, which deletes the pod, marks the execution CANCELLED and
returns early, so nothing would ever poll it again -- this is the most common
Tangle-initiated deletion, and without an explicit release here every
cancellation would wedge a pod;
- the SYSTEM_ERROR handler in the running-executions queue, which terminalizes
an execution the poll loop could not process.

And a backstop for everything else (the orchestrator restarting mid-flight, an
execution row that no longer exists, a pod whose node was deleted so it never
reaches a terminal phase): `release_abandoned_finalizer_holds` force-releases
any held pod whose deletion has been pending longer than
`finalizer_hold_max_duration` (default 10 minutes), swept at most once per
`finalizer_sweep_interval` (default 60s) from the running-executions queue. The
worst case is therefore bounded by configuration rather than by correctness of
the release paths.

Deliberate choices worth knowing about:

- Off by default. A hold that is never released is worse than a pod collected
too early. Enable per launcher with `hold_pods_until_status_observed=True`, or
with `CLOUD_PIPELINES_HOLD_PODS_UNTIL_STATUS_OBSERVED=true` for deployments
that build the launcher without threading the argument through.
- Pods only, never Jobs. A Job outlives its pods, so the Job launcher does not
need this, and a finalizer on Job-owned pods would wedge the Job controller's
own cleanup. `_KubernetesPodOrJobLauncher` composes rather than inherits, so
it forwards the sweep explicitly and passes the hold to the pod launcher only.
- The finalizer is added after the pod post-processor runs and before the pod is
created, so a post-processor cannot drop it and the pod is never unprotected.
- Removal uses a JSON Patch with a `test` op. `metadata.finalizers` has merge
semantics, so a strategic merge patch can add a finalizer but never remove
one; the `test` op makes removal atomic against a concurrent writer, and a
conflict is retried.
- The sweep lists pods cluster-wide, because a pod post-processor may place pods
in a namespace other than the launcher's own. That needs `list pods` at
cluster scope; without it the sweep logs and returns, and the per-execution
releases still work.
- Releasing is always best-effort at the call sites: the execution's outcome is
already committed, so a failure to release must never change it.

This protects status and exit code. Logs survive only as long as the node does:
if the node is gone, `read_namespaced_pod_log` fails whether or not the pod
object is held, which is why log upload stays best-effort.

Co-authored-by: Morgan Wowk [morgan.wowk@shopify.com](mailto:morgan.wowk@shopify.com)

### DIFF
--- a/cloud_pipelines_backend/launchers/interfaces.py
+++ b/cloud_pipelines_backend/launchers/interfaces.py
@@ -137,6 +137,21 @@ class LaunchedContainer(abc.ABC):
     def terminate(self) -> None:
         raise NotImplementedError()
 
+    def release(self) -> bool:
+        """Releases any hold the orchestrator has on the launched container.
+
+        A launcher may keep the underlying workload object alive after the
+        container has finished, so that its terminal status, exit code and logs
+        can still be observed (the Kubernetes pod launcher does this with a
+        finalizer). Calling `release` says "I have everything I need" and lets
+        the cluster reclaim the object.
+
+        Returns True when a hold was actually released. Launchers that hold
+        nothing return False, which is the default, so callers can call this
+        unconditionally.
+        """
+        return False
+
 
 # @dataclasses.dataclass
 # class ContainerExecutionResult:
--- a/cloud_pipelines_backend/launchers/kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/kubernetes_launchers.py
@@ -35,6 +35,97 @@ _MAIN_CONTAINER_NAME = "main"
 # Kubernetes annotation keys. (Has strict naming policy. Single slash only etc.)
 _CLOUD_PIPELINES_KUBERNETES_ANNOTATION_KEY = "cloud-pipelines.net"
 _KUBERNETES_LAUNCHER_ANNOTATION_KEY = "cloud-pipelines.net/launchers.kubernetes"
+# Finalizer that keeps a task pod's API object alive after the container has
+# finished, until the orchestrator has observed its terminal status. While it is
+# present a delete only stamps `deletionTimestamp`: the object stays readable in
+# `Terminating` instead of disappearing mid-poll. See `LaunchedContainer.release`.
+_STATUS_OBSERVED_FINALIZER = "cloud-pipelines.net/status-observed"
+# Finalizers are not selectable, so the hold is mirrored in a label to make held
+# pods findable by `release_abandoned_finalizer_holds`.
+_STATUS_OBSERVED_FINALIZER_LABEL_KEY = "cloud-pipelines.net/held-for-status"
+# Opt-in, because a hold that is never released is worse than a pod that is
+# collected too early. Set on the launcher, or through this environment variable
+# for deployments that construct the launcher without threading the argument.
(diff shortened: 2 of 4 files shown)


morgan-wowk:
* **#317** <a href="https://app.graphite.com/github/pr/TangleML/tangle/317?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/317?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Ark-kun:
Good idea.
Although extra care is needed (you already do this) so that we do not abandon the holds, which adds complexity.
We need to decide best course of action when SYSTEM_ERROR occurs. Especially, give that releasing the finalizer can result in another error if Kubernetes API Server is unresponsive.

@morgan-wowk 
Note: I've recently added `LaunchedContainer.cleanup()`. Some of your `.release()` code might become part of or superseded by `cleanup()`. https://github.com/TangleML/tangle/pull/318

Issue #139, opened 2026-03-06
bug: Launchers - Kubernetes - Multi-node - System error when getting pod logs due to Job failure when some pods are still initializing


Ark-kun:
```
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"container \"main\" in pod \"tangle-ce-019cc4bebb3318802061-1-26cmm\" is waiting to start: PodInitializing","reason":"BadRequest","code":400}
```
```
Traceback (most recent call last):
  File "/app/backend/cloud_pipelines_backend/orchestrator_sql.py", line 165, in internal_process_running_executions_queue
    self.internal_process_one_running_execution(
  File "/app/backend/cloud_pipelines_backend/orchestrator_sql.py", line 896, in internal_process_one_running_execution
    _retry(reloaded_launched_container.upload_log)
  File "/app/backend/cloud_pipelines_backend/orchestrator_sql.py", line 1030, in _retry
    return func()
           ^^^^^^
  File "/app/backend/cloud_pipelines_backend/launchers/kubernetes_launchers.py", line 1466, in upload_log
    all_logs = self._get_all_logs()
               ^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/cloud_pipelines_backend/launchers/kubernetes_launchers.py", line 1439, in _get_all_logs
    pod_key: self._get_log_by_pod_key(pod.metadata.name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/cloud_pipelines_backend/launchers/kubernetes_launchers.py", line 1428, in _get_log_by_pod_key
    log = core_api_client.read_namespaced_pod_log(
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/kubernetes/client/api/core_v1_api.py", line 24264, in read_namespaced_pod_log
    return self.read_namespaced_pod_log_with_http_info(name, namespace, **kwargs)  # noqa: E501
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/kubernetes/client/api/core_v1_api.py", line 24387, in read_namespaced_pod_log_with_http_info
[...]

PR #325, opened 2026-08-12
fix(orchestrator): losing pod logs must not fail an execution

morgan-wowk:
## What breaks today

When the GKE cluster-autoscaler evicts and deletes a Job's pod mid-run, the Job object still reports `Failed`, so the orchestrator has everything it needs to record a clean `FAILED`. Today it doesn't:

1. `LaunchedKubernetesJob._get_log_by_pod_key` re-raises on a `404` — only a `400` ("Bad Request", pod still initializing) was mapped to "no logs".
2. The orchestrator's `FAILED` branch calls `upload_log()` unguarded, so that raise propagates. `upload_log` reads logs by remembered pod name → the pod is gone → `404` → `_retry` burns all 5 attempts and re-raises → the outer handler overwrites `FAILED` with `SYSTEM_ERROR`.

Observed twice on one production run: `Exception calling <bound method LaunchedKubernetesJob.upload_log ...>` ×5, then `Error processing running container execution`.

## New rules

- **A log-upload failure never changes an execution's terminal status.** The `FAILED` branch now guards `upload_log` exactly as the `SUCCEEDED` branch already does — a failed upload is logged and the execution stays `FAILED` with normal downstream skipping.
- **A vanished pod means "no logs", not an error.** `_get_log_by_pod_key` returns `None` on a `404` (pod deleted), just as it already does on a `400` (pod initializing), with a warning naming the pod. `_debug_pods` deliberately retains pods that no longer exist, so a deleted pod would otherwise re-`404` on every later read. See [issue #139](https://github.com/TangleML/tangle/issues/139) for the original `400` case.

## Why both fixes ship as one unit

Each fix alone leaves the failure reachable: the orchestrator guard stops the `FAILED → SYSTEM_ERROR` flip for *any* upload failure, while the `404` handling stops the vanished-pod read from raising in the first place — so logs from other still-present pods in the same Job still upload. Together they give one guarantee: losing logs never changes a terminal status.

## Scope

- `LaunchedKubernetesContainer.get_log` / `stream_log_lines` (the single-pod launcher) are left unchanged. They currently raise on any `ApiException`, and the orchestrator guard already protects the `FAILED` path for both launcher types; extending them would change pod-launcher semantics, so it's deliberately out of scope.
- `podFailurePolicy` / `backoffLimitPerIndex` / `maxFailedIndexes` are untouched — a separate change.

## Tests

- `tests/test_kubernetes_launcher_error_classification.py`: `_get_log_by_pod_key` returns `None` on a `404` and still raises on `403` and `500` (mocked `CoreV1Api`).
- `tests/test_orchestrator_failed_log_upload.py`: an execution whose launcher raises from `upload_log` ends `FAILED` (not `SYSTEM_ERROR`), with downstream marked `SKIPPED`.
- Full suite green (`uv run pytest`).

## Downstream

[Shopify/oasis-backend](https://github.com/Shopify/oasis-backend) pins this repo as the `backend` submodule; it needs a submodule pointer bump to pick this fix up.

### DIFF
--- a/cloud_pipelines_backend/launchers/kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/kubernetes_launchers.py
@@ -1563,6 +1563,15 @@ class LaunchedKubernetesJob(interfaces.LaunchedContainer):
                 # Kubernetes client raises kubernetes.client.exceptions.ApiException when Pod is still in PodInitializing phase
                 # See https://github.com/TangleML/tangle/issues/139
                 return None
+            if ex.status == http.HTTPStatus.NOT_FOUND:
+                # The Pod is gone (e.g. deleted by the cluster-autoscaler mid-run).
+                # `_debug_pods` deliberately retains vanished Pods, so this key would
+                # 404 on every subsequent read. A deleted Pod means "no logs", not an
+                # error, so return None instead of re-raising.
+                _logger.warning(
+                    f"Pod {pod_name} no longer exists; its logs are unrecoverable."
+                )
+                return None
             raise
 
     def _get_all_logs(self) -> dict[str, str]:
--- a/cloud_pipelines_backend/orchestrator_sql.py
+++ b/cloud_pipelines_backend/orchestrator_sql.py
@@ -1065,7 +1065,16 @@ class OrchestratorService_Sql:
                     message=orchestration_error_message,
                 )
 
-            _retry(reloaded_launched_container.upload_log)
+            # A log-upload failure must not change a terminal status.
+            # Logs are best-effort; losing them is never a reason to fail an
+            # execution. The launcher decides what a lost log means; the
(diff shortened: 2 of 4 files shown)


morgan-wowk:
* **#326** <a href="https://app.graphite.com/github/pr/TangleML/tangle/326?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#325** <a href="https://app.graphite.com/github/pr/TangleML/tangle/325?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/325?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Ark-kun (approved):

PR #330, opened 2026-08-14
Bound orchestrator storage waits

morgan-wowk:
## What

- Add `GoogleCloudStorageProviderWithTimeout` — a GCS `StorageProvider` that passes an explicit per-request `timeout=` to every GCS call (upload / download / `exists` / `get_info` / `list_blobs`). The timeout is configurable via `TANGLE_GCS_REQUEST_TIMEOUT_SECONDS` (default `60s`).
- Wire the Kubernetes launchers (`kubernetes_launchers`, `google_kubernetes_launchers`) to construct and use it.
- Bound the orchestrator retry loop with a wall-clock deadline (`TANGLE_ORCHESTRATOR_RETRY_DEADLINE_SECONDS`, default `180s`) via a new `RetryDeadlineExceededError`, so retries can no longer spin unbounded.
- Document every storage-provider function with the concrete Tangle use cases it serves (docstrings only; no behavior change).
- Add tests for the timeout provider and the retry deadline.

## Why / system impact

Before: GCS calls relied on the client's default (effectively unbounded) timeout. A single hung GCS request could wedge the orchestrator's poll loop indefinitely, stalling all execution processing.

After: every storage wait is bounded. A slow or stuck GCS request now fails fast and is retried within a deadline instead of blocking the loop, so one bad request degrades a single execution rather than the whole orchestrator.

## Paired change

A consumer opts in by constructing this provider instead of the upstream `GoogleCloudStorageProvider`. The paired consumer change is linked in a follow-up comment.

### DIFF
--- a/cloud_pipelines_backend/launchers/google_kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/google_kubernetes_launchers.py
@@ -2,7 +2,7 @@ import typing
 
 from kubernetes import client as k8s_client_lib
 
-from cloud_pipelines.orchestration.storage_providers import google_cloud_storage
+from cloud_pipelines_backend.storage_providers import google_cloud_storage_with_timeout
 
 from . import kubernetes_launchers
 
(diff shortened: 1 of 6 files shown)


morgan-wowk:
Paired consumer change (opts into the new timeout-bounded provider): Shopify/oasis-backend#507 → https://github.com/Shopify/oasis-backend/pull/507

morgan-wowk:
Closing in favor of a metrics-first approach. Before tuning any timeouts, we want to measure actual GCS upload/download durations on prod to find where to focus. The stock google-cloud-storage client already enforces a 60s per-request timeout + ~120s retry deadline, so a per-request override at the same default was largely a no-op; the real questions need data first. New instrumentation PR incoming.

PR #333, opened 2026-08-17
Trace GCS upload/download duration and bytes

morgan-wowk:
## Why

A recent incident had Tangle pipelines stalling on GCS transfers. The first fix attempt was per-request timeouts ([#330](https://github.com/TangleML/tangle/pull/330), now closed), but the stock `google-cloud-storage` client already enforces a 60s per-request timeout plus a ~120s retry deadline, so a same-default override added little. Before choosing any timeout policy we need to know **how long real uploads/downloads take on prod** — so we can find concretely where the time goes and which operation to focus on.

## Out of scope

GCS Fuse operations, which is the more impactful area to measure, is not covered by this change.

## Changes

- New `cloud_pipelines_backend/instrumentation/gcs_tracing.py`: `TracingGoogleCloudStorageProvider`, a subclass of the upstream `GoogleCloudStorageProvider` that overrides only the four public transfer methods (`upload`, `upload_bytes`, `download`, `download_bytes`). Each wraps `super()` in an OpenTelemetry span.
- The GKE launchers now construct the tracing provider instead of the plain one (one-line swap at each site; the positional `gcs_client` argument is unchanged).

Each span uses the `tangle.storage` tracer, is named `gcs.<operation>`, and carries:
- `gcs.operation` — `upload` / `upload_bytes` / `download` / `download_bytes`
- `gcs.uri` — the object/prefix being transferred
- `gcs.bytes` — payload size, recorded **only** for the in-memory transfers (`upload_bytes` / `download_bytes`, where it is a free `len(data)`)

On failure the span is marked `ERROR` and the exception is recorded, then re-raised unchanged.

**Duration is the primary signal** and is intrinsic to every span. We deliberately do **not** walk the local filesystem to size file/dir transfers: that scan is strictly dominated by the transfer it would measure (the base provider already does one GCS round-trip per file), it added a robustness hole (a `stat` race could fail an otherwise-successful transfer), and authoritative sizes for those artifacts already live in `artifact_data.total_size` and can be joined offline.

## System / UX impact

- **Measurement only** — no transfer behavior changes; `__init__` is not overridden and no timeouts are introduced.
- Adds one span per GCS transfer (exported via the already-configured OTLP `BatchSpanProcessor`). No extra filesystem access.

## Before / after

- **Before:** GCS transfer latency was invisible; we could only guess where time went.
- **After:** every launcher-side upload/download emits a span with its duration, queryable in Observe by operation and URI.

## Follow-up

Companion oasis-backend PR bumps the submodule and swaps the orchestrator-side provider so orchestrator-initiated GCS ops (`exists`, `get_info`, small `download_bytes`, output preservation) are covered too.

### DIFF
--- a/cloud_pipelines_backend/instrumentation/gcs_tracing.py
+++ b/cloud_pipelines_backend/instrumentation/gcs_tracing.py
@@ -1,0 +1,64 @@
+"""GCS storage provider that emits one OTel span per upload/download so transfer duration is measurable."""
+
+from __future__ import annotations
+
+import collections.abc
+import contextlib
+
+from opentelemetry import trace
+from opentelemetry.trace import StatusCode
+
+from cloud_pipelines.orchestration.storage_providers import google_cloud_storage
+
+_tracer = trace.get_tracer("tangle.storage")
+
+
+@contextlib.contextmanager
+def _gcs_operation_span(
+    operation: str, uri: str
+) -> collections.abc.Iterator[trace.Span]:
+    with _tracer.start_as_current_span(
+        f"gcs.{operation}",
+        attributes={"gcs.operation": operation, "gcs.uri": uri},
+    ) as span:
+        try:
+            yield span
+        except Exception as exception:
+            span.set_status(StatusCode.ERROR)
+            span.record_exception(exception)
+            raise
+
+
+class TracingGoogleCloudStorageProvider(
+    google_cloud_storage.GoogleCloudStorageProvider
+):
+    def upload(
+        self,
+        source_path: str,
+        destination_uri: google_cloud_storage.GoogleCloudStorageUri,
+    ) -> None:
+        with _gcs_operation_span("upload", destination_uri.uri):
+            super().upload(source_path, destination_uri)
+
+    def upload_bytes(
+        self, data: bytes, destination_uri: google_cloud_storage.GoogleCloudStorageUri
+    ) -> None:
+        with _gcs_operation_span("upload_bytes", destination_uri.uri) as span:
+            span.set_attribute("gcs.bytes", len(data))
+            super().upload_bytes(data, destination_uri)
+
+    def download(
+        self,
+        source_uri: google_cloud_storage.GoogleCloudStorageUri,
+        destination_path: str,
+    ) -> None:
+        with _gcs_operation_span("download", source_uri.uri):
+            super().download(source_uri, destination_path)
+
+    def download_bytes(
+        self, source_uri: google_cloud_storage.GoogleCloudStorageUri
+    ) -> bytes:
+        with _gcs_operation_span("download_bytes", source_uri.uri) as span:
+            data = super().download_bytes(source_uri)
+            span.set_attribute("gcs.bytes", len(data))
+            return data
--- a/cloud_pipelines_backend/launchers/google_kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/google_kubernetes_launchers.py
@@ -2,9 +2,8 @@ import typing
 
 from kubernetes import client as k8s_client_lib
 
-from cloud_pipelines.orchestration.storage_providers import google_cloud_storage
-
 from . import kubernetes_launchers
+from ..instrumentation import gcs_tracing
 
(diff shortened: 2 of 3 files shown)


morgan-wowk:
* **#333** <a href="https://app.graphite.com/github/pr/TangleML/tangle/333?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TangleML/tangle/333?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

cloud_pipelines_backend/instrumentation/gcs_tracing.py
```diff
@@ -0,0 +1,79 @@
...
+def _local_path_size_bytes(path: str) -> int:
+    if os.path.isdir(path):
+        total = 0
+        for root, _dirs, files in os.walk(path):
+            for name in files:
+                file_path = os.path.join(root, name)
+                if not os.path.islink(file_path):
+                    total += os.path.getsize(file_path)
+        return total
+    return os.path.getsize(path)
```
morgan-wowk: Considering removing this. Doing some scaling analysis now. My thought is that this type of filetree walking and size scanning will not scale to thousands of files.

morgan-wowk: Asking Claude now how many files we would realistically be dealing with in this loop.

morgan-wowk: **About to apply the following change:**

> Keep gcs.bytes = len(data) on upload_bytes / download_bytes — O(1), exact, already in memory.
Remove _local_path_size_bytes and the gcs.bytes set on upload / download. For those, rely on the span duration (your primary signal) plus artifact_data.total_size when you want size correlation.
Net result: zero filesystem walking, no robustness hole, duration everywhere, exact bytes on the small in-memory transfers, and directory/file sizes still available authoritatively from artifact_data.

morgan-wowk: A file tree walk has been removed

cloud_pipelines_backend/instrumentation/gcs_tracing.py
```diff
@@ -0,0 +1,64 @@
...
+) -> collections.abc.Iterator[trace.Span]:
+    with _tracer.start_as_current_span(
+        f"gcs.{operation}",
+        attributes={"gcs.operation": operation, "gcs.uri": uri},
+    ) as span:
+        try:
+            yield span
+        except Exception as exception:
+            span.set_status(StatusCode.ERROR)
+            span.record_exception(exception)
```
Mbeaulne, line 28: **AI review:** `start_as_current_span()` already records escaping exceptions and marks the span as `ERROR` by default. Because this handler also calls `set_status()` and `record_exception()` before re-raising, each failed transfer emits two exception events (confirmed with an in-memory span exporter). Could we remove the manual exception handling, or explicitly disable the context manager’s automatic handling?

Mbeaulne (approved):
AI review left some comments. If they are true, fix. Otherwise I've approved

camielvs (approved):
I don't have enough python & backend context to review on an implementation level, but overall the intent looks fine to me

morgan-wowk:
Closing in favor of orchestrator-side telemetry.

The inv-26772 investigation showed the orchestrator freeze happens in `internal_process_one_running_execution`, where `get_reader().get_info()` walks every file of a large Directory artifact on the single orchestrator thread. That is **not** on the path this PR instruments — the provider `upload`/`download`/`upload_bytes`/`download_bytes` methods here are the small control-plane transfers; the bulk artifact I/O goes through gcsfuse mounts, and the file-tree walk that actually blocks the loop goes through `get_info()`, which these overrides do not wrap.

Superseded by a stack that instruments the actual freeze path, attributed by `container_execution.id`:
- https://github.com/TangleML/tangle/pull/334 — span around per-execution processing
- https://github.com/TangleML/tangle/pull/335 — span around the `get_info()` output-resolution hot path

PR #341, opened 2026-08-20
feat: Orchestrator - Launchers - Kubernetes - Start labelling created Pods, Jobs and Services with a Tangle label

Ark-kun:
This is useful to quickly identify objects created by Tangle

### DIFF
--- a/cloud_pipelines_backend/launchers/kubernetes_launchers.py
+++ b/cloud_pipelines_backend/launchers/kubernetes_launchers.py
@@ -36,7 +36,18 @@ _MAIN_CONTAINER_NAME = "main"
 
 # Kubernetes annotation keys. (Has strict naming policy. Single slash only etc.)
 _CLOUD_PIPELINES_KUBERNETES_ANNOTATION_KEY = "cloud-pipelines.net"
+_TANGLE_KUBERNETES_LABEL_KEY = "tangle.tangleml.com"
 _KUBERNETES_LAUNCHER_ANNOTATION_KEY = "cloud-pipelines.net/launchers.kubernetes"
+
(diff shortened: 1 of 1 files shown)


cloud_pipelines_backend/launchers/kubernetes_launchers.py
```diff
@@ -36,7 +36,18 @@
 
 # Kubernetes annotation keys. (Has strict naming policy. Single slash only etc.)
 _CLOUD_PIPELINES_KUBERNETES_ANNOTATION_KEY = "cloud-pipelines.net"
+_TANGLE_KUBERNETES_LABEL_KEY = "tangle.tangleml.com"
```
morgan-wowk, line 39: Question: Would you like me to refactor https://github.com/Shopify/oasis-backend/pull/458 to use this label instead of `tangleml.com` or refactor this PR to use `tangleml.com` without the subdomain?

morgan-wowk: The default namespace is already created for staging but not yet prod on e09.

Ark-kun: TBH, I do not have an opinion here.  
The main reason I added the `tangle.` prefix was to distinguish Tangle pods from Tangent pods.  
I also thought about recent pushes to make Tangent "bigger" than Tangle (although those were scaled down by senior leadership).  
  
I'd say, just proceed with [Shopify/oasis-backend#458](https://github.com/Shopify/oasis-backend/pull/458) as-is. We are already using the `tangleml.com` in `launchers.py`. We can refactor all of that later.

